PR #8, opened 2024-08-13
More off-chain changes

naddison36:
* Added `upgradeProxy` Hardhat task

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,6 +57,16 @@ deploy-holesky:
 	@forge script script/001_OETH_ARM.s.sol --rpc-url $(HOLESKY_URL) --private-key ${DEPLOYER_PRIVATE_KEY} --broadcast --slow --verify -vvvv
 #	@forge script script/deploy/DeployManager.sol --rpc-url $(HOLESKY_URL) --private-key ${HOLESKY_DEPLOYER_PRIVATE_KEY} --broadcast --slow --verify -vvvv
 
+# Upgrade scripts
+upgrade:
+	@forge script script/002_Upgrade.s.sol --rpc-url $(PROVIDER_URL) --private-key ${DEPLOYER_PRIVATE_KEY} --broadcast --slow --verify -vvvv
+
+upgrade-testnet:
+	@forge script script/002_Upgrade.s.sol --rpc-url $(TESTNET_URL) --private-key ${DEPLOYER_PRIVATE_KEY} --broadcast --slow -vvvv
+
+upgrade-holesky:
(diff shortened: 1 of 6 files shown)


naddison36:
#### Authentication

- [ ] Never use tx.origin
- [ ] Every external/public function is supposed to be externally accessible
- [ ] Every external/public function has the correct authentication

#### Ethereum

- [ ] Contract does not send or receive Ethereum.
- [ ] Contract has no payable methods.
- [ ] Contract is not vulnerable to being sent self destruct ETH

#### Cryptographic code


#### Gas problems

- [ ] Contracts with for loops must have either
- [ ] Contracts with for loops must not allow end users to add unlimited items to a loop that is used by others or admins.

#### Black magic

- [ ] Does not contain `selfdestruct`
- [ ] Does not use `delegatecall` outside of proxying. _If an implementation contract were to call delegatecall under attacker control, it could call selfdestruct the implementation contract, leading to calls through the proxy silently succeeding, even though they were failing._
- [ ] Address.isContract should be treated as if could return anything at any time, because that's reality.

#### Overflow

- [ ] Code is solidity version >= 0.8.0
- [ ] All for loops use uint256

#### Proxy

- [ ] No storage variable initialized at definition when contract used as a proxy implementation.

#### Events
- [ ] All state changing functions emit events

## Medium Checks

#### Rounding and casts

- [ ] Contract rounds in the protocols favor
- [ ] Contract does not have bugs from loosing rounding precision
- [ ] Code correctly multiplies before division
- [ ] Contract does not have bugs from zero or near zero amounts
- [ ] Safecast is aways used when casting

#### Dependencies

- [ ] Review any new contract dependencies thoroughly (e.g. OpenZeppelin imports) when new dependencies are added or version of dependencies changes.
- [ ] If OpenZeppelin ACL roles are use review & enumerate all of them.
- [ ] Check OpenZeppelin [security vulnerabilities](https://github.com/OpenZeppelin/openzeppelin-contracts/security/advisories) and see if any apply to current PR considering the version of OpenZeppelin contract used.

#### External calls

- [ ] Contract addresses passed in are validated
- [ ] No unsafe external calls
- [ ] Reentrancy guards on all state changing functions
- [ ] Oracles, one of:
  - [ ] No oracles (in new code)
  - [ ] Oracles can't be bent
  - [ ] If oracle can be bent, it won't hurt us.
- [ ] Do not call balanceOf for external contracts to determine what they will do when they use internal accounting

#### Deploy

- [ ] Deployer permissions are removed after deploy

#### Strategy checks

- [ ] Check balance cannot be manipulated up AND down by an attacker
- [ ] No read only reentrancy on downstream protocols during checkBalance

#### Logic


#### Deployment Considerations


#### Attack


#### Flavor

PR #16, opened 2024-09-18
Refactor test structure

clement-ux:


lib/forge-std
naddison36, line 1: should we be git ignoring the `lib` folder?

naddison36 (approved):

lib/forge-std
clement-ux: No, I don't think this is necessary, as lib only contains folders from gitmodules.
But if you prefer I can do it.

PR #17, opened 2024-09-19
Switch from gitmodules to Soldeer.

clement-ux:
This PR aims to switch from gitmodules to soldeer for managing dependencies.

> Soldeer is a package manager for Solidity built in Rust.

As mentioned on the Soldeer [repo](https://github.com/mario-eth/soldeer/blob/main/README.md):
> - git submodules in foundry are not a good solution for managing dependencies
> - npmjs was built for the js ecosystem not for solidity
> - github versioning of the releases is a pain and not all the projects are using it correctly

To ensure all the dependencies have been added run:
```
make install
```
This will:
- update foundry
- install all dependencies from soldeer
- install all dependencies from yarn

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -32,4 +32,10 @@ remappings_regenerate = false
 remappings_prefix = "@"
 remappings_location = "config"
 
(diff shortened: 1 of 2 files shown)


naddison36 (approved):
Looks good

PR #31, opened 2024-10-07
Add Zapper for LidoARM

clement-ux:
## General

Zapper to deposit liquidity in ARM with raw ETH instead of wETH in a single transaction.

## Constraint for `claimRedeem`

At the moment, as only the msg.sender can be the withdrawer, it is not possible for the zapper to withdraw for someone and then unwrap wETH for ETH.

### DIFF
--- a/src/contracts/Interfaces.sol
+++ b/src/contracts/Interfaces.sol
@@ -107,7 +107,7 @@ interface IOethARM {
     function claimWithdrawals(uint256[] calldata requestIds) external;
 }
 
-interface ILiquidityProviderARM {
+interface ILiquidityProviderARM is IERC20 {
     function previewDeposit(uint256 assets) external returns (uint256 shares);
     function deposit(uint256 assets) external returns (uint256 shares);
 
@@ -256,7 +256,3 @@ interface IStETHWithdrawal {
     function getLastRequestId() external view returns (uint256);
 }
 
-interface ILidoARM {
-    function deposit(uint256 assets) external returns (uint256 shares);
-    function transfer(address to, uint256 shares) external returns (bool);
-}
--- a/src/contracts/Interfaces.sol
+++ b/src/contracts/Interfaces.sol
@@ -255,3 +255,8 @@ interface IStETHWithdrawal {
     function getWithdrawalRequests(address _owner) external view returns (uint256[] memory requestsIds);
     function getLastRequestId() external view returns (uint256);
 }
+
+interface ILidoARM {
+    function deposit(uint256 assets) external returns (uint256 shares);
+    function transfer(address to, uint256 shares) external returns (bool);
+}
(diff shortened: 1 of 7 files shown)


src/contracts/ZapperLidoARM.sol
```diff
@@ -0,0 +1,51 @@
...
+// Interfaces
+import {IWETH} from "./Interfaces.sol";
+import {IERC20} from "./Interfaces.sol";
+import {ILidoARM} from "./Interfaces.sol";
+
+contract ZapperLidoARM is Ownable {
+    IWETH public immutable weth;
+    ILidoARM public immutable lidoArm;
+
+    event Zap(address indexed sender, uint256 shares);
```
naddison36: I think we should also include the amount of ETH (assets) deposited

src/contracts/Interfaces.sol
```diff
@@ -255,3 +255,8 @@ interface IStETHWithdrawal {
     function getWithdrawalRequests(address _owner) external view returns (uint256[] memory requestsIds);
     function getLastRequestId() external view returns (uint256);
 }
+
+interface ILidoARM {
```
naddison36: There's a ILiquidityProviderARM interface that already has deposit. We can extend IERC20 so we also get transfer.

src/contracts/ZapperLidoARM.sol
```diff
@@ -0,0 +1,51 @@
...
+
+// Contracts
+import {Ownable} from "./Ownable.sol";
+
+// Interfaces
+import {IWETH} from "./Interfaces.sol";
+import {IERC20} from "./Interfaces.sol";
+import {ILidoARM} from "./Interfaces.sol";
+
+contract ZapperLidoARM is Ownable {
```
naddison36, line 17: I don't think we need the Zapper to be `Ownable`. I see that is used for `rescueERC20` but there shouldn't be any tokens in this contract. I see the OETH Zapper doesn't have this.

If we do want the Zapper to be Ownerable, we need to decide who the owner should be. Options are:
1. 5/8 multisig
2. Strategist 2/9 multisig
3. ARM Defender Relayer
My preference would be the Strategist

@DanielVF what do you think?

src/contracts/ZapperLidoARM.sol
naddison36: I've pushed the change to add the amount of assets (WETH) deposited

naddison36 (approved):

src/contracts/ZapperLidoARM.sol
clement-ux: I completely agree, it was just in case someone mistakenly send token to it (like people sending wETH to wETH contract). 
I have no problem removing it.

However if we want to keep it, I think Strategist or Relayer is enough.

PR #32, opened 2024-10-09
Add Zap compliant with Liquidity Provider Cap

clement-ux:
## Current issue
The current zap implementation allows users to bypass checks on the Liquidity Provider Cap. 

## How can we fix it
We can add a new function that can be called only by the zap contract, with the liquidity provider address as an argument.

By modifying a bit the logic of deposit function, we can do 2 extra things:
- send shares directly to the liquidity provider (instead of msg.sender)
- check the liquidity provider cap of the liquidity provider (instead of msg.sender)

## Downside
This increases the complexity of the deposit logic, and once the `accountCapEnabled` is disabled, this complexity will not be necessary anymore (except for gas optimization). So maybe it's not worth it to add it now and add a simple Zapper once `accountCapEnabled` is disabled.


naddison36 (approved):
LGTM

src/contracts/AbstractARM.sol
```diff
@@ -432,17 +451,17 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
 
         // mint shares
-        _mint(msg.sender, shares);
+        _mint(liquidityProvider, shares);
 
         // Add the deposited assets to the last available assets
         lastAvailableAssets += SafeCast.toInt128(SafeCast.toInt256(assets));
 
         // Check the liquidity provider caps after the new assets have been deposited
         if (capManager != address(0)) {
```
naddison36, line 460: I've just realised we can avoid the extra call to capManager if we also check the msg.sender is Zapper.
I'll change no the `nicka/lp` branch as this has already been merged

PR #40, opened 2024-10-25
N-03 Use calldata instead of memory for external function parameters

naddison36:
# Changes

* The `requestLidoWithdrawals` and `claimLidoWithdrawals` functions in `LidoARM` have been changed to use `calldata`


DanielVF (approved):

clement-ux (approved):
LGTM 👍

PR #51, opened 2024-11-25
Allow setCrossPrice to be called by Operator

naddison36:
* Currently only the Owner can call `setCrossPrice` which is a 5/8 Safe. The allows the Operator to also call `setCrossPrice`. There are already protections in `setCrossPrice` so the cross price can not be lowered when there is  stETH in the ARM. This prevents stETH being sold for less than it was purchased for.


clement-ux (approved):
LGTM 👍

PR #52, opened 2024-11-26
Handle donation attack of Lido withdrawal NFTs

naddison36:
# Changes

* Someone can temporarily freeze withdrawals from the Lido queue by transferring a Lido withdrawal NFT and then claiming it via the Lido ARM. As the withdrawal request was not done through the Lido ARM, the `lidoWithdrawalQueue` amount will be smaller than the ETH received by the claim process. This will cause and underflow if all withdrawal requests are claimed.
* The avoid underflows, `lidoWithdrawalQueue` is set to zero if its going to underflow.


naddison36:
Closing for an alternative approach of storing the withdrawal request amount so any Lido mass slashing events can be better socialised to LP holders.

PR #53, opened 2024-12-04
Track the Lido withdrawal request amounts

naddison36:


clement-ux (approved):
Nothing to say, everything looks good  👍

DanielVF (approved):
We'll need to ensure we have no active withdraws when this goes live.

PR #59, opened 2024-12-16
Fix setFee when changing from a zero performance fee

naddison36:


clement-ux (approved):
I ran the invariants for 30min and nothing breaks. 
Looks good to me 👍

DanielVF (approved):

PR #71, opened 2025-04-24
Oracle Router

clement-ux:
## Description
- Add an Oracle router for harvester.


clement-ux:
After [this](https://www.notion.so/originprotocol/What-Oracle-to-use-for-Silo-rewards-1e584d46f53c80d0bef2d588b64d7f0d?pvs=4) design decision, oracle is not needed.

PR #79, opened 2025-05-05
ARM integration to Lending Markets

naddison36:
# Design Specification

The following spec is copied from https://www.notion.so/originprotocol/ARM-Lending-Markets-Design-1d684d46f53c805984b3f9b4c1cbadc8

The design decisions can be exported on request https://www.notion.so/originprotocol/ARM-Lending-Markets-Design-1d684d46f53c805984b3f9b4c1cbadc8?p=1cf84d46f53c80b8ad3ef07f6a45dd12&pm=s

## Features

- An ARM maintains a liquidity buffer in the ARM to buy LSTs at a discount. Excess liquidity is deposited into the active lending market to earn yield.
- There can be many support lending markets but only one can be used at any point in time.
- The ARM will integrate to the lending markets using ERC-4626.

## ARM Lending Market Use Cases

- Any can allocate liquidity in the ARM. This tries to adjust the ARM’s liquidity to match the ARM’s liquidity buffer.
    - This requires an active lending market to be set. A lending market with a zero address will disable the lending market feature.
    - If not enough liquidity in the ARM
        - and there is enough liquidity in the active lending market
            - liquidity is withdrawn from the active lending market.
        - and there is not enough enough liquidity in the active lending market
            - all liquidity is redeemed from the active lending market.
    - If too much liquidity in the ARM, liquidity is deposited to the active lending market.
- The Owner (Timelock) can add a supported lending market
    - The ARM’s liquidity asset must match the asset of the lending market’s vault
- The Owner (Timelock) can remove a supported lending market
    - Must not be the active lending market
- The Operator (Defender Relayer) can switch the active lending market
    - The lending market has to be one of the ARM’s supported lending markets
    - All liquidity is withdrawn from the previous active lending market.
        - If not all the liquidity can be removed as the market’s utilization is high, the transaction will fail
    - Liquidity is deposited to the new active lending market via the allocate process
- The Owner (Timelock) can set the ARM liquidity buffer

## Harvester Use Cases

The Harvesters on Sonic will use the Magpie aggregator to get best execution of swapping Silo for wS.

Some key design decisions

- Anyone can `collect` token rewards from the supported lending market strategies. For the Silo lending markets, that will be the Silo token. Rewards from multiple lending market strategies can be collected before being swapped to the ARM’s associated liquidity asset.
- Only the Owner or Operator can `swap` token rewards for the associated ARM’s liquidity asset. For the Silo lending markets, that’s swapping Silo for wrapped S (wS). The initial implementation will only support the Magpie DEX aggregator. The Owner or Operator is responsible for calling the Magpie APIs and passing the the Magpie specific swap data.
- The Harvester will do the following validators of swaps
    - Swaps are to the associated ARM’s liquidity asset. ie wS
    - The swap price is within an allowed slippage from the Oracle price
    - The receiver of the swap is the Harvester’s rewards recipient. Initially that will be the ARM contract. That could be changed to a Dripper in the future.
- There will be a separate Harvester contract for each ARM. This is a lot simplifier than having a single Harvester that has to account for the incoming rewards tokens and sending the appropriate liquidity assets to each ARM.

# Contracts

* Change license from MIT to BUSL-1.1
* Upgraded `AbstractARM` to park idle liquidity in a lending market
* Added `OriginARM` that implements `AbstractARM` by integrating to an Origin Vault
* Added `SiloMarket` contract that can harvest Silo rewards from a Silo lending market
* Added `Harvester` that can collect and swap Silo rewards to wS using the Magpie DEX aggregator
* Added `ZapperARM` that integrates to multiple ARMs by swapping S to wS.

sonicContracts

# Testing

To run all the tests

```bash
make test
```

To run the OriginARM fork tests

```bash
make test-c-Fork_Concrete_OriginARM
```

# Deployment

The deployment scripts are
*`script/deploy/sonic/001_DeployOriginARMProxy.sol`
*`script/deploy/sonic/002_DeployOriginARM.sol`

In the `.env` file, set the `DEPLOYER_PRIVATE_KEY` and `SONIC_URL` variables

```bash
make deploy-sonic
```


naddison36:
## Requirements

See the PR description or Notion https://www.notion.so/originprotocol/ARM-Lending-Markets-Design-1d684d46f53c805984b3f9b4c1cbadc8

## Easy Checks

#### Authentication

- [x] Never use tx.origin
- [x] Every external/public function is supposed to be externally accessible
- [x] Every external/public function has the correct authentication

#### Ethereum

- [x] Contract does not send or receive Ethereum.
   `ZapperARM` receives native S but that's ok.
- [x] Contract has no payable methods.
   `deposit` on `ZapperARM` is payable but that's ok.
- [x] Contract is not vulnerable to being sent self destruct ETH

#### Cryptographic code

- [x] This contract code does not roll it's own crypto.
- [x] No signature checks without reverting on a 0x00 result.
- [x] No signed data could be used in a replay attack, on our contract or others.

#### Gas problems

- [x] Contracts with for loops must have either:
    - [ ] A way to remove items
    - [ ] Can be upgraded to get unstuck
    - [ ] Size can only controlled by admins
- [x] Contracts with for loops must not allow end users to add unlimited items to a loop that is used by others or admins.

#### Black magic

- [x] Does not contain `selfdestruct`
- [x] Does not use `delegatecall` outside of proxying. _If an implementation contract were to call delegatecall under attacker control, it could call selfdestruct the implementation contract, leading to calls through the proxy silently succeeding, even though they were failing._
   Only the existing `Proxy` contract uses `delegatecall`
- [x] Address.isContract should be treated as if could return anything at any time, because that's reality.

#### Overflow

- [x] Code is solidity version >= 0.8.0
- [x] All for loops use uint256

#### License

- [x] The contract uses the appropriate limited BUSL-1.1 (Business) or the open MIT license
- [ ] If the contract license changes from MIT to BUSL-1.1 any contracts importing it need to also have their license set to BUSL-1.1

#### Proxy

- [x] No storage variable initialized at definition when contract used as a proxy implementation.

#### Events

- [x] All state changing functions emit events
  The ERC-4626 functions on SiloMarkets currently do not emit events

## Medium Checks

#### Rounding and casts
- [ ] Contract rounds in the protocols favor
- [ ] Contract does not have bugs from loosing rounding precision
- [ ] Code correctly multiplies before division
- [ ] Contract does not have bugs from zero or near zero amounts
- [ ] Safecast is aways used when casting

#### Dependencies
- [ ] Review any new contract dependencies thoroughly (e.g. OpenZeppelin imports) when new dependencies are added or version of dependencies changes.
- [ ] If OpenZeppelin ACL roles are use review & enumerate all of them.
- [ ] Check OpenZeppelin [security vulnerabilities](https://github.com/OpenZeppelin/openzeppelin-contracts/security/advisories) and see if any apply to current PR considering the version of OpenZeppelin contract used.

#### External calls

- [ ] Contract addresses passed in are validated
- [ ] No unsafe external calls
- [ ] Reentrancy guards on all state changing functions
    - [ ] Still doesn't protect against external contracts changing the state of the world if they are called.
- [ ] No malicious behaviors
- [ ] Low level call() must require success.
- [ ] No slippage attacks (we need to validate expected tokens received)
- [ ] Oracles, one of:
  - [ ] No oracles
  - [ ] Oracles can't be bent
  - [ ] If oracle can be bent, it won't hurt us.
- [ ] Do not call balanceOf for external contracts to determine what they will do when they use internal accounting

#### Tests

  - [ ] Each publicly callable method has a test
  - [ ] Each logical branch has a test
  - [ ] Each require() has a test
  - [ ] Edge conditions are tested
  - [ ] If tests interact with AMM make sure enough edge cases (pool tilts) are tested. Ideally with fuzzing.

#### Deploy

- [ ] Deployer permissions are removed after deploy

## Strategy Specific

_Remove this section if the code being reviewed is not a strategy._

#### Strategy checks

- [ ] Check balance cannot be manipulated up AND down by an attacker
- [ ] No read only reentrancy on downstream protocols during checkBalance
- [ ] All reward tokens are collected
- [ ] The harvester can sell all reward tokens
- [ ] No funds are left in the contract that should not be as a result of depositing or withdrawing
- [ ] All funds can be recovered from the strategy by some combination of depositAll, withdraw, or withdrawAll()
- [ ] WithdrawAll() can always withdraw an amount equal to or larger than checkBalances report, even in spite of attacker manipulation.
- [ ] WithdrawAll() cannot be MEV'd
- [ ] WithdrawAll() does not revert when strategy has 0 assets
- [ ] Strategist cannot steal funds

#### Downstream

- [ ] We have monitoring on all backend protocol's governances
- [ ] We have monitoring on a pauses in all downstream systems

## Thinking

#### Logic

  _Are there bugs in the logic?_

  - [ ] Correct usage of global & local variables. -> they might differentiate only by an underscore that can be overlooked (e.g. address vs _address).

#### Deployment Considerations

_Are there things that must be done on deploy, or in the wider ecosystem for this code to work. Are they done?_

#### Internal State

- What can be always said about relationships between stored state
- What must hold true about state before a function can run correctly (preconditions)
- What must hold true about the return or any changes to state after a function has run.

Does this code do that?

#### Attack

_What could the impacts of code failure in this code be._

_What conditions could cause this code to fail if they were not true._

_Does this code successfully block all attacks._

#### Flavor

_Could this code be simpler?_

_Could this code be less vulnerable to other code behaving weirdly?_

src/contracts/AbstractARM.sol
```diff
@@ -184,7 +205,9 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
 
         // Initialize the last available assets to the current available assets
         // This ensures no performance fee is accrued when the performance fee is calculated when the fee is set
-        lastAvailableAssets = SafeCast.toInt128(SafeCast.toInt256(_availableAssets()));
+        _setFee(_fee);
```
DanielVF, line 208: 🟠 Is there a reason _setFee is called twice?

src/contracts/AbstractARM.sol
```diff
@@ -112,7 +114,14 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
     /// @notice The address of the CapManager contract used to manage the ARM's liquidity provider and total assets caps.
     address public capManager;
 
-    uint256[41] private _gap;
+    /// @notice The address of the active lending market.
+    address public activeMarket;
```
DanielVF, line 118: Code might be prettier if `activeMarket` were typed as a IERC4626, instead of needing to be casted to an IERC4626 when it's used.

src/contracts/AbstractARM.sol
```diff
@@ -598,26 +642,35 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
+    function _availableAssets() internal view returns (uint256 availableAssets, uint256 outstandingWithdrawals) {
+        // Liquidity assets, eg WETH, in the ARM and lending markets are priced at 1.0
         // Base assets, eg stETH, in the withdrawal queue are also priced at 1.0
         // Base assets, eg stETH, in the ARM are priced at the cross price which is a discounted price
         uint256 assets = IERC20(liquidityAsset).balanceOf(address(this)) + _externalWithdrawQueue()
             + IERC20(baseAsset).balanceOf(address(this)) * crossPrice / PRICE_SCALE;
 
-        // The amount of liquidity assets (WETH) that is still to be claimed in the withdrawal queue
-        uint256 outstandingWithdrawals = withdrawsQueued - withdrawsClaimed;
+        if (activeMarket != address(0)) {
```
DanielVF, line 654: ⛽️ `activeMarket` will be read from storage three times here in normal use. Gas savings to making it a temp variable.

src/contracts/AbstractARM.sol
```diff
@@ -726,4 +771,132 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
+        address previousActiveMarket = activeMarket;
+        require(previousActiveMarket != _market, "ARM: already active market");
+
+        if (previousActiveMarket != address(0)) {
+            // Redeem all shares from the previous active lending market.
+            // balanceOf is used instead of maxRedeem to ensure all shares are redeemed.
+            // maxRedeem can return a smaller amount of shares than balanceOf if the market is highly utilized.
+            uint256 shares = IERC4626(previousActiveMarket).balanceOf(address(this));
+            // This could fail if the market has high utilization
+            if (shares > minSharesToRedeem) {
```
DanielVF, line 821: If we don't do a redeem here because of minSharesToRedeem, we'll lose accounting on these funds and count it as a loss.

Presumably that's okay, since minSharesToRedeem will be a very small number, but there should be some kind of not here that this is intentional.

src/contracts/AbstractARM.sol
```diff
@@ -726,4 +771,132 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
+        (uint256 availableAssets, uint256 outstandingWithdrawals) = _availableAssets();
+        if (availableAssets == 0) return;
+
+        int256 armLiquidity = SafeCast.toInt256(IERC20(liquidityAsset).balanceOf(address(this)))
+            - SafeCast.toInt256(outstandingWithdrawals);
+        uint256 targetArmLiquidity = availableAssets * armBuffer / 1e18;
+
+        int256 liquidityDelta = armLiquidity - SafeCast.toInt256(targetArmLiquidity);
+
+        if (liquidityDelta > 0) {
```
DanielVF, line 852: Perhaps we should have some kind of hysteresis on `liquidityDelta`, rather just allocating when it's not perfect.

Well behaved 4626's round against the protocol, so it feels like it would be be possible to get into a situation where someone could call allocate() over and over again, each time removing a rounding error of value.

Just feels a little safer and more efficient in general to have something like a 1% threshold required on either side.

src/contracts/AbstractARM.sol
```diff
@@ -726,4 +771,132 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
         fees = SafeCast.toUint256(assetIncrease) * fee / FEE_SCALE;
     }
+
+    ////////////////////////////////////////////////////
+    ///         Lending Market Functions
+    ////////////////////////////////////////////////////
+
+    /// @notice Owner adds supported lending market to the ARM.
+    /// @param _markets The addresses of the lending markets to add
+    function addMarkets(address[] calldata _markets) external onlyOwner {
```
DanielVF, line 781: In order to be a safe market for the ARM, it must be:
- up only exchange rate
- no slippage
- no fees.

Don't know if we need to note the properties required anywhere.

src/contracts/Harvester.sol
```diff
@@ -0,0 +1,267 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {Initializable} from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import {IERC20, IERC20Metadata} from "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
+import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
```
DanielVF, line 7: Do we use this `Math` anywhere?

src/contracts/Harvester.sol
```diff
@@ -0,0 +1,267 @@
...
+        onlyOperatorOrOwner
+        returns (uint256 toAssetAmount)
+    {
+        uint256 recipientAssetBefore = IERC20(liquidityAsset).balanceOf(rewardRecipient);
+
+        // Validate the swap data and do the swap
+        toAssetAmount = _doSwap(swapPlatform, fromAsset, fromAssetAmount, data);
+
+        // Check the recipient of the swap got the reported amount of liquidity assets
+        uint256 recipientAssets = IERC20(liquidityAsset).balanceOf(rewardRecipient) - recipientAssetBefore;
```
DanielVF, line 127: Note that if an attacker got code execution at any point during the swap, then an attacker could _deposit_ into the vault, thus properly increasing the vaults funds, and defeating this check that the swap happened.

Fortunately, the oracle check is on `toAssetAmount`, which would check the swap results of an honest swap system.

The checks here fail if:
- We have no oracle, and attacker gets execution during a swap
- Malicious swap contract.

Given that this code would not catch malicious swap contract, it makes me wonder if we should just trust what is returned from the swap contract and skip the complexity here around the separate balance checks.

src/contracts/Harvester.sol
```diff
@@ -0,0 +1,267 @@
...
+        if (recipientAssets < toAssetAmount) {
+            revert BalanceMismatchAfterSwap(recipientAssets, toAssetAmount);
+        }
+
+        emit RewardTokenSwapped(fromAsset, liquidityAsset, swapPlatform, fromAssetAmount, toAssetAmount);
+
+        // If there is no price provider, we exit early
+        if (priceProvider == address(0)) return toAssetAmount;
```
DanielVF, line 135: 🔴 If no oracle, and unless there's some magical anti-mev system in place, this swap method will get sandwich attacked to death when swapping

Are we really planning on doing no oracle swaps with no passed in expected minimum amount?

src/contracts/Harvester.sol
```diff
@@ -0,0 +1,267 @@
...
+import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
+
+import {OwnableOperable} from "./OwnableOperable.sol";
+import {IHarvestable, IMagpieRouter, IOracle} from "./Interfaces.sol";
+
+/**
+ * @title Collects rewards from strategies and swaps them for ARM liquidity tokens.
+ * @author Origin Protocol Inc
+ */
+contract Harvester is Initializable, OwnableOperable {
```
DanielVF, line 16: Maybe we should give this contract a slightly different name, since we already have a quite different "Harvester" contract already.

src/contracts/Harvester.sol
```diff
@@ -0,0 +1,267 @@
...
+        if (IERC20Metadata(_liquidityAsset).decimals() != 18) revert InvalidDecimals();
+
+        liquidityAsset = _liquidityAsset;
+        magpieRouter = _magpieRouter;
+    }
+
+    /// @notice
+    function initialize(address _priceProvider, uint256 _allowedSlippageBps, address _rewardRecipient)
+        external
+        initializer
```
DanielVF, line 77: I like it slightly better when initialize has onlyOwner on it. This matters less now that we usually initialize during the same transaction as the proxy upgrade.

src/contracts/Harvester.sol
DanielVF: Or maybe this is passed in the datablob.

src/contracts/AbstractARM.sol
naddison36: It's either declare the storage variable as `IERC4626` can cast to `address` when needed. eg param validation or token approve. Or declare as `address` and cast to `IERC4626`

activeMarket  is used as `IERC4626` nine times and `address` eight times.

I prefer to cast to `IERC4626` when needed as it makes it clear the functions belong to `IERC4626`

src/contracts/AbstractARM.sol
naddison36: This must be a merge issue. I'll remove the duplicate.

src/contracts/AbstractARM.sol
naddison36: Good pickup. I'll change as `AbstractARM` will also be used for the Lido ARM on Ethereum

src/contracts/Harvester.sol
naddison36: no, I'll remove. Thanks

src/contracts/Harvester.sol
naddison36: We can parse the `amountOutMin` from the Magpie data and check against the allowed slippage. But if the Operator is compromised, the attacker can just set the `amountOutMin` within the allowed slippage of the bad swap price.

src/contracts/Harvester.sol
naddison36: Good point. I'll rename to `SonicHarvester` as it has Sonic specific logic. eg Magpie DEX aggregator

src/contracts/Harvester.sol
naddison36: I'll add `onlyOwner` to `initialize`

src/contracts/Harvester.sol
naddison36: An interesting point that the attacker could get execution control during the swap. That's possible if the swap is multi part and involves transferring native S to the attacker, or a swap is done against an AMM that has a callback.

We had this swap balance check in the Ethereum Swapper that used 1Inch.

I suspect if this check was ever triggered it would be from a rounding issue that we wouldn't be that concerned about anyway.

@clement-ux would you be able to run the tests that compared the gas usage with the `BalanceMismatchAfterSwap` check and without?

clement-ux: >@clement-ux would you be able to run the tests that compared the gas usage with the BalanceMismatchAfterSwap check and without?

@naddison36  Yes sure!

clement-ux: ## Test
- Removing the following section of the code.
Capture d’écran 2025-05-12 à 12 11 12


## Result
| With | Without | Diff |
| - | - | - |
| 242135 | 240348 | 0.74% |


## FlameChart and FlameGraph
image

image

naddison36: thanks @clement-ux 

So not a big difference which probably comes down to the reads are to warmed storage slots.

I'm inclined to leave wth `BalanceMismatchAfterSwap` check in.

@DanielVF what do you think?

src/contracts/AbstractARM.sol
naddison36: This is a good point. We can have something like

```
if (liquidityDelta > minLiquidityThreshold) {
```

I was thinking we could reuse `minSharesToRedeem` or `MIN_TOTAL_SUPPLY` but those are of type `uint256`. And its better to not mix concerns.

naddison36: I've added `allocateThreshold` immutable in PR https://github.com/OriginProtocol/arm-oeth/pull/84

```Solidity
// The allocateThreshold prevents the ARM from constantly depositing and withdrawing if there are rounding issues
if (liquidityDelta > allocateThreshold) {
    // We have too much liquidity in the ARM, we need to deposit some to the active lending market

    uint256 depositAmount = SafeCast.toUint256(liquidityDelta);

    IERC20(liquidityAsset).approve(activeMarket, depositAmount);
    IERC4626(activeMarket).deposit(depositAmount, address(this));
} else if (liquidityDelta < 0) {
```

src/contracts/AbstractARM.sol
naddison36: I've added this comment to the Natspec

src/contracts/AbstractARM.sol
DanielVF: image

I am fine with that.

src/contracts/markets/SiloMarket.sol
```diff
@@ -0,0 +1,192 @@
+// SPDX-License-Identifier: BUSL-1.1
```
DanielVF, line 1: The reasons we have this separate wrapper contract is to avoid needing to put `collectRewards` and `setHarvester` onto into the OriginArm contract? Does this have any other differences than just using the 4626 directly?

src/contracts/OriginARM.sol
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * @notice Claim multiple previously requested withdrawals from the Origin Vault.
+     * The caller should check the withdrawal has passed the withdrawal delay
+     * and there is enough liquidity in the Vualt.
+     * @param requestIds The request IDs of the withdrawal requests.
+     * @param amountClaimed The total amount claimed across all withdrawal requests.
+     */
+    function claimOriginWithdrawals(uint256[] calldata requestIds) external returns (uint256 amountClaimed) {
```
DanielVF, line 85: Why do we run out of liquidity? Mostly when we are the best price on the market. Which means that there might be room to charge more money. 

This suggests that in an ideal world, whenever we re-added liquidity, from whatever source, be it moving funds from a lending platform, Lido withdraws, or claiming from a vault, that we would want this to be done after potential price change. In order to enforce this, we would need to have these funds moving in functions only be called by the owner/operator.

DanielVF: I don't think there's any harm in running it the current way. We certainly can choose to do so. it's just there's a possibility of trading getting off earning more money in exchange for more work on the backend side.

src/contracts/LidoARM.sol
```diff
@@ -38,7 +38,7 @@ contract LidoARM is Initializable, AbstractARM {
     /// @param _lidoWithdrawalQueue The address of the Lido's withdrawal queue contract
     /// @param _claimDelay The delay in seconds before a user can claim a redeem from the request
     constructor(address _steth, address _weth, address _lidoWithdrawalQueue, uint256 _claimDelay)
-        AbstractARM(_weth, _steth, _weth, _claimDelay)
+        AbstractARM(_weth, _steth, _weth, _claimDelay, 0)
```
DanielVF, line 41: Should we have a non-zero value here since this is going to be hard-coded into the next deploy?

src/contracts/Harvester.sol
DanielVF: It's not going to hurt anything having the extra check, outside 2,000 gas and a little extra code. I'm okay with keeping it if you want.

DanielVF:
> The lending market integration was made a lot simpler by only having one active lending market. If there is a large amount of idle liquidity deposited into that active market, then that can diminish the returns as the utilization of the market will drop. It could be more profitable to split the idle liquidity across multiple lending markets but that adds a lot of complexity that isn't needed for now.

I agree it's good to run just one active market for now.

src/contracts/AbstractARM.sol
```diff
@@ -726,4 +772,136 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
+        address previousActiveMarket = activeMarket;
+        require(previousActiveMarket != _market, "ARM: already active market");
+
+        if (previousActiveMarket != address(0)) {
+            // Redeem all shares from the previous active lending market.
+            // balanceOf is used instead of maxRedeem to ensure all shares are redeemed.
+            // maxRedeem can return a smaller amount of shares than balanceOf if the market is highly utilized.
+            uint256 shares = IERC4626(previousActiveMarket).balanceOf(address(this));
+            // This could fail if the market has high utilization
+            if (shares > minSharesToRedeem) {
```
DanielVF, line 826: This if statement breaks our up only invariant and the defense beyond that to a loss of user funds has to do with gas prices and amount of funds in our contract. While we're probably good there, it's not something I like very much.

What's the harm in removing this if in `setActiveMarket`? Worse case is that we have to send a tiny amount of shares to unlock a switch from an almost empty lending market? This could be done by the operator, or by allocating more in.

(The minSharesToRedeem if statement in `_allocate` is fine, since it doesn't affect accounting if it runs or not)

DanielVF: It would still be fine to have an if shares greater than zero check here.

src/contracts/markets/SiloMarket.sol
```diff
@@ -0,0 +1,192 @@
...
+
+        emit Withdraw(arm, arm, arm, assets, shares);
+    }
+
+    /// @notice Get the amount of asset tokens that can be received
+    /// from burning an exact amount of Silo lending market shares.
+    /// @return assets The amount of asset tokens that will be received.
+    function previewRedeem(uint256 shares) external view returns (uint256 assets) {
+        // Preview the amount of assets that can be redeemed for a given number of shares
+        assets = IERC4626(market).previewRedeem(shares);
```
DanielVF, line 123: I wonder if it's worth checking that reentrancyGuardEntered is false, so that we don't have to worry about any kind of read only reentrancy.

image

src/contracts/AbstractARM.sol
```diff
@@ -112,7 +114,14 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
 
-    uint256[41] private _gap;
+    /// @notice The address of the active lending market.
+    address public activeMarket;
+    /// @notice Lending markets that can be used by the ARM.
+    mapping(address market => bool supported) public supportedMarkets;
+    /// @notice Percentage of liquid assets to keep in the ARM. 100% = 1e18.
+    uint256 public armBuffer;
+
+    uint256[39] private _gap;
```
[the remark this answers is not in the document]
naddison36: This has been fixed in https://github.com/OriginProtocol/arm-oeth/pull/79/commits/997eb6db8926e93b336d18d781d8574b7ae66698

src/contracts/markets/SiloMarket.sol
naddison36: That's correct. We only have `SiloMarket` so the Harvester can collect rewards. The 4626 functions just pass the calls through to the Silo vault which is also 4626.

src/contracts/LidoARM.sol
naddison36: The `minSharesToRedeem` was introduced to get around an issue with the Silo vaults which failed if a small amount of shares was redeemed. eg 100 wei of shares

We'll have to see if the lending market used by LidoARM will have the same problem.

src/contracts/OriginARM.sol
```diff
@@ -0,0 +1,102 @@
...
+     * @param requestIds The request IDs of the withdrawal requests.
+     * @param amountClaimed The total amount claimed across all withdrawal requests.
+     * @return amountClaimed The total amount of oTokens claimed from the Origin Vault.
+     */
+    function claimOriginWithdrawals(uint256[] calldata requestIds) external returns (uint256 amountClaimed) {
+        // Claim the previously requested withdrawals from the Origin Vault.
+        (, amountClaimed) = IOriginVault(vault).claimWithdrawals(requestIds);
+
+        // Store the reduced amount outstanding withdrawals from the Origin Vault
+        vaultWithdrawalAmount -= amountClaimed;
```
pandadefi, line 91: In case a withdrawal request is transferred to the OriginARM contract, you won't be able to claim it because this line will underflow.

pandadefi: This issue also exists in the lido contract.

https://github.com/OriginProtocol/arm-oeth/blob/2e4cffa8548cbd860bdcae9c64a5cbd9a79b2145/src/contracts/LidoARM.sol#L155

A simple fix could be
```
if (vaultWithdrawalAmount <  amountClaimed) {
vaultWithdrawalAmount = 0;
}
```

src/contracts/markets/SiloMarket.sol
DanielVF: Bumping this. Silo is crawling with hooks. In theory they've got everything nicely covered and correctly ordered, but I would much rather be checking this for our silo integration.

src/contracts/AbstractARM.sol
```diff
@@ -550,6 +574,14 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
         // Store the updated claimed amount
         withdrawsClaimed += SafeCast.toUint128(assets);
 
+        // If there is not enough liquidity assets in the ARM, get from the active market
```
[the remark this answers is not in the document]
naddison36: Changed in https://github.com/OriginProtocol/arm-oeth/pull/101/commits/cc88ff57dcc6bfd77c8876ab40c6362e437580a6

naddison36:
@DanielVF looking at example `deposit` and `allocate` txs for read-only reentrancy, the calls are contained within the Silo contracts.
There are a lot of calls from the Varlamore S Vault as it loops over the different markets and calls the various Silo lending market contracts. But none of them call out to third party contracts. eg token contracts.
There are also no native S transfers in the Silo calls.

Deposit
49f8b659

Allocate
55574c2a

pandadefi (approved):

PR #84, opened 2025-05-13
Added allocateThreshold

naddison36:


src/contracts/AbstractARM.sol
```diff
@@ -850,7 +859,8 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
 
         int256 liquidityDelta = armLiquidity - SafeCast.toInt256(targetArmLiquidity);
 
-        if (liquidityDelta > 0) {
+        // The allocateThreshold prevents the ARM from constantly depositing and withdrawing if there are rounding issues
+        if (liquidityDelta > allocateThreshold) {
```
DanielVF, line 863: There's are a couple way this could be done:

- absolute value (current implementation): ex. Need to deposit at least $10,000 before allocating
- fixed percentage of armFunds: ex. Need to deposit at least 1% of arm funds over threshold before allocating

I would tend to lean towards the second, but I see that would increase the gas costs. The current implementation is definitely better than not having anything, and I think is fine.

DanielVF (approved):

PR #103, opened 2025-05-30
Return the liquidity delta from allocate to help with off-chain processing

naddison36:


clement-ux (approved):
All good, looks good to me 👍

PR #109, opened 2025-07-22
Add OS Price Script

shahthepro:
- Adds a new price script for OS ARM that uses lending APY and market price to figure out an optimal buy price to be set on the ARM contract

TODO:
- Figure out a way to deploy it


clement-ux (approved):
Looks good to me 👍

PR #114, opened 2025-08-07
Add lending market data to `snapLido`

clement-ux:
## Description

This pull request updates the `logAssets` function in `src/js/tasks/lido.js` to include balances and percentages for assets held in the active lending market. The changes add support for fetching and displaying the lending market balance and its proportion of total assets, using a hardcoded market address for now.

**Lending Market Integration:**

* Added logic to retrieve the ARM's balance in the active lending market (`Abstract4626MarketWrapper`), calculate its value in assets, and include it in the total asset calculation. The market address is currently hardcoded but will be dynamic in the future.
* Updated output to display the lending market balance and its percentage share alongside other asset categories.

(I generated description using copilot, not bad)

### DIFF
--- a/src/js/actions/setPrices.js
+++ b/src/js/actions/setPrices.js
@@ -30,13 +30,14 @@ const handler = async (event) => {
       // buyPrice: 0.9997,
       maxSellPrice: 0.9999,
       minSellPrice: 0.9998,
-      maxBuyPrice: 0.9980,
-      minBuyPrice: 0.9950,
+      maxBuyPrice: 0.998,
+      minBuyPrice: 0.995,
       inch: true,
       amount: 100,
       tolerance: 0.2,
       fee: 0.8,
       offset: 0.4,
+      priceOffset: true,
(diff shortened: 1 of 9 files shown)


src/js/actions/setPrices.js
```diff
@@ -30,13 +30,13 @@ const handler = async (event) => {
...
+      maxBuyPrice: 0.9980,
+      minBuyPrice: 0.9950,
+      inch: true,
       amount: 100,
-      tolerance: 0.1,
+      tolerance: 0.2,
       fee: 0.8,
-      offset: 0.2,
+      offset: 0.4,
       blockTag: "latest",
```
naddison36, line 40: lets add given this is our new default
`priceOffset: true,`

naddison36 (approved):

src/js/actions/setPrices.js
naddison36: I've done this change

PR #161, opened 2025-11-25
Improve Linter

clement-ux:
# Description
Forge provides a new `linter` based on [Solar](https://github.com/paradigmxyz/solar), a Rust based solidity compiler.
This PR aims to discuss about should we use it or not?
Note: we can choose to only run it on contracts, not on script or test.

## Pros
- We could have a standardized approach for variable naming (this is the biggest point). 
- We could enforce it when creating a PR, to ensure code is standardized.
- It could help us to fix issue before they appear: like `unsafe-typecast` (this is the only case I encountered):
```
warning[unsafe-typecast]: typecasts that can truncate values should be checked
   --> src/contracts/AbstractARM.sol:594:33
    |
594 |         uint40 claimTimestamp = uint40(block.timestamp + claimDelay);
    |                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
 ```

## Cons
- Less flexibility on the naming (but this help to standardize the code).
- A lot of false alarm like this one, but we can easily remove it with inline comment though.
```
warning[erc20-unchecked-transfer]: ERC20 'transfer' and 'transferFrom' calls should check the return value
   --> src/contracts/AbstractARM.sol:651:9
    |
651 |         IERC20(liquidityAsset).transfer(msg.sender, assets);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
 ```

# How to try
Ensure you use last Foundry version + build
```
foundryup
forge b
```
Try linter
```
forge l
```

### Forge-lint rules — by severity

| Severity               | Lint ID / Rule               | What it checks / Risk |
|------------------------|-----------------------------|-----------------------|
| **High**               | `incorrect-shift`            | Bit-shift where a literal is shifted by a variable — likely a wrong operand order / logic bug. |
|                        | `unchecked-call`             | Low-level calls (`.call()`, `.delegatecall()`, `.staticcall()`) without checking success return — risk of silent failure / security issue. |
|                        | `erc20-unchecked-transfer`   | ERC-20 `transfer` / `transferFrom` calls ignoring the boolean return — transfers might silently fail. |
| **Medium**             | `divide-before-multiply`     | Integer arithmetic doing division before multiplication in the same expression — may cause precision loss (due to truncation). |
|                        | `unsafe-typecast`            | Unsafe type casts (e.g. large → smaller integer) — risk of data loss or unexpected behaviour. |
| **Info / Style**       | `pascal-case-struct`         | Enforces `PascalCase` naming for structs — style / readability. |
|                        | `mixed-case-function`        | Enforces `mixedCase` (camelCase) for function names — style / readability. |
|                        | `mixed-case-variable`        | Enforces `mixedCase` naming for mutable variables — style / readability. |
|                        | `screaming-snake-case-const` | Enforces `SCREAMING_SNAKE_CASE` naming for `constant` variables — style / readability. |
|                        | `screaming-snake-case-immutable` | Enforces `SCREAMING_SNAKE_CASE` naming for `immutable` variables — style / readability. |
|                        | `unused-import`              | Warns about imports that are never used — avoids unnecessary imports / clutter. |
|                        | `unaliased-plain-import`     | Warns when using plain imports without alias — encourages clearer import naming / avoids name conflicts. |
| **Gas optimizations**  | `asm-keccak256`              | Suggests using inline-assembly `keccak256` when appropriate — may reduce gas usage vs high-level call. |
| **Code-size (bytecode)** | `unwrapped-modifier-logic` | Flags modifiers containing logic outside the `_` placeholder — such logic is duplicated per function, increasing contract size. |

- Here is the list of all lints that are supported, ranked by severity: https://getfoundry.sh/forge/linting/#erc20-unchecked-transfer
- Here is the global documentation: https://getfoundry.sh/config/reference/linter/#linter-configuration


clement-ux:
Closing this experimental branch

PR #203, opened 2026-04-07
Reduced the Ethena ARM DELAY_REQUEST from 3 hours to 30 minutes

naddison36:
Ethena recently reduced sUSDe's cooldown from 7 days to just 1 day. This means the Ethena ARM can request a lot more withdrawals in a day using its 42 EthenaUnstaker contracts.

PR #204, opened 2026-04-07
build: migrate JS tooling from yarn to pnpm

clement-ux:
## Summary

Migrates the JavaScript dependency manager from **yarn** (classic v1) to **pnpm** for faster installs, smaller `node_modules` via the content-addressable store, and stricter dependency hoisting.

### pnpm migration
- **`Makefile`** — `yarn install` → `pnpm install`; `clean-all` now removes `pnpm-lock.yaml`
- **`package.json`** — added `"packageManager": "pnpm@9.15.0"` so Corepack picks the right version automatically
- **`.github/actions/setup/action.yml`** — replaced manual yarn cache step with the standard pnpm CI pattern: `pnpm/action-setup@v4` + `actions/setup-node@v4` (Node 20, `cache: 'pnpm'`); renamed input `install-yarn` → `install-pnpm`. The action reads the pnpm version directly from `package.json#packageManager`
- **`.github/workflows/main.yml`** — 4 jobs updated to `install-pnpm: 'true'`
- **`CLAUDE.md`** / **`.claude/CLAUDE.md`** — doc references updated
- **Lockfile** — `yarn.lock` deleted, `pnpm-lock.yaml` committed

### Test fixes (unrelated pre-existing CI failures, fixed here to get the PR green)
- **Smoke test** — cherry-picked `13fb74d` from #203: updates `morphoMarket` reference and adjusts the WETH allocation in `test_allocate_to_lending_market` to account for outstanding withdrawals, so the assertion holds against current mainnet state.
- **Fork test (FlyTrade)** — when the upstream FlyTrade API returns Cloudflare error 525 (origin SSL handshake failure), the `flyTradeQuote` hardhat task now emits a sentinel hex (`0xc10ad11ae525`) and exits 0 instead of crashing. `Helpers.getFlyTradeQuote` detects that sentinel from `vm.ffi` output and calls `vm.skip(true)`, so transient upstream outages mark the test as skipped rather than failing CI. **Only 525 triggers the skip** — every other failure mode still surfaces as a real test failure.

## Notes

- pnpm is stricter than yarn classic about phantom dependencies. Local install + lint + rollup all pass cleanly, so nothing was relying on undeclared transitive deps.
- `dependencies/@pendle-sy-1.0.0-1.0.0/yarn.lock` is left untouched (it belongs to a vendored Soldeer dep).
- pnpm 9 requires Node ≥ 18.12. The CI setup action pins Node 20.

## Test plan

- [x] `pnpm install` succeeds and produces `pnpm-lock.yaml`
- [x] `pnpm lint` passes
- [x] `pnpm prettier:check` passes
- [x] `pnpm rollup -c src/js/actions/rollup.config.cjs` bundles all Defender Actions
- [ ] CI: `build`, `unit-tests`, `fork-tests`, `smoke-tests` jobs install pnpm and pass


.github/actions/setup/action.yml
```diff
@@ -43,17 +43,20 @@ runs:
...
+      uses: pnpm/action-setup@v4
+
+    - name: Setup Node.js
+      if: inputs.install-pnpm == 'true'
+      uses: actions/setup-node@v4
       with:
-        path: node_modules
-        key: yarn-${{ hashFiles('yarn.lock') }}
-        restore-keys: yarn-
+        node-version: '20'
```
naddison36: I think we should be using node 22. Node 20 is on its last version and security updates end this month.

naddison36 (dismissed):

.github/actions/setup/action.yml
clement-ux: Fixed in this commit: [37aee94](https://github.com/OriginProtocol/arm-oeth/pull/204/commits/37aee94e7f9dbfa83b5db484f17f46b10e23d817)

naddison36 (approved):
I've tested my common Hardhat tasks and they are working.

PR #208, opened 2026-04-10
New ARM features before Ethena audit

naddison36:
## Summary

Adds the next ARM upgrade set ahead of the Ethena audit. This PR now includes active-market liquidity sourcing during swaps, discounted base-asset swap fees, per-price liquidity limits, multi-base-asset support, adapter-based redemption flows, updated loss/fee accounting, pause controls, reentrancy protection, and safer upgrade/migration handling.

## Merged Changes

- [#221 Multiple base assets](https://github.com/OriginProtocol/arm-oeth/pull/221)
- [#223 Pro-rata losses to redeemers and remaining](https://github.com/OriginProtocol/arm-oeth/pull/223)
- [#224 Calculate swap fee against cross price rather than par value](https://github.com/OriginProtocol/arm-oeth/pull/224)
- [#225 Value base assets in the base withdrawal queue at the cross price](https://github.com/OriginProtocol/arm-oeth/pull/225)
- [#226 Added pause and unpause](https://github.com/OriginProtocol/arm-oeth/pull/226)
- [#227 Reentrancy protection on swaps](https://github.com/OriginProtocol/arm-oeth/pull/227)
- [#231 Custom errors for Operator and Admin functions](https://github.com/OriginProtocol/arm-oeth/pull/231)
- [#232 Support claiming legacy redeem requests](https://github.com/OriginProtocol/arm-oeth/pull/232)
- [#234 Fix token approval in adapters](https://github.com/OriginProtocol/arm-oeth/pull/234)
- [#235 Adds legacy withdrawal queue safety checks and selector documentation](https://github.com/OriginProtocol/arm-oeth/pull/235)
- [#236 Adapter Natspec and layout](https://github.com/OriginProtocol/arm-oeth/pull/236)
- [#244 Restore slot storage of Ethena ARM](https://github.com/OriginProtocol/arm-oeth/pull/244)
- [#245 Replace stored pending unstaker index array with total request counter](https://github.com/OriginProtocol/arm-oeth/pull/245)
- [#246 Sweep donated ETH and WETH from claim adapters](https://github.com/OriginProtocol/arm-oeth/pull/246)
- [#250 Fix legacy withdraw queue migration initializer version](https://github.com/OriginProtocol/arm-oeth/pull/250)
- [#253 Drop migrateLegacyWithdrawQueue, gate upgrades with per-ARM check](https://github.com/OriginProtocol/arm-oeth/pull/253)

### Withdraw From Active Market During Swaps

Swaps that need more liquidity than is currently held in the ARM can now pull the shortfall from the active ERC-4626 lending market.

This lets the ARM keep more liquidity deployed while still supporting swaps, provided the active market has enough withdrawable liquidity.

### Discount-Based Swap Fees

Replaces the old performance fee on asset growth with fees accrued when the ARM buys base assets at a discount.

Fees are calculated using a swap fee multiplier and measured against the configured cross price, aligning fee accrual with the ARM's valuation price for each base asset.

### Per-Price Liquidity Limits

Buy and sell liquidity limits are tracked per configured price and emitted in traderate updates.

Operators can cap how much liquidity can be consumed at the current buy/sell prices before prices are refreshed. Swap-side liquidity checks now cover both buy and sell paths.

### Multi-Base-Asset ARM

Moves reusable multi-base logic into `AbstractARM`, allowing a single ARM to support multiple base assets against one liquidity asset.

Examples:

- Lido ARM: `stETH`, `wstETH`
- EtherFi ARM: `eETH`, `weETH`
- Ethena ARM: `sUSDe`
- Origin ARM: `OETH`, `wOETH`

Each base asset has its own config:

- buy price
- sell price
- buy liquidity remaining
- sell liquidity remaining
- cross price
- pending redeem assets
- pegged/non-pegged conversion mode
- redemption adapter

### Asset Adapter Redemption Flow

Protocol-specific redemption logic has moved out of the ARM and into per-asset adapters.

Adapters now own details such as:

- Lido withdrawal NFT queue
- wstETH unwrap and Lido withdrawal request
- EtherFi withdrawal NFTs
- Ethena cooldown/unstaker flow
- Origin vault withdrawal requests
- wrapped Origin asset redemption

The ARM tracks generic pending redeem accounting in liquidity-asset terms. Lido and EtherFi-style claim adapters also sweep adapter-held ETH/WETH when claiming, so donated balances do not remain stranded.

### LP Loss Accounting

Updates LP redeem accounting so losses are shared pro-rata between redeemers and remaining LPs instead of allowing one side to absorb the full impact.

If assets per share falls after a redeem request is created, the claim uses the lower claim-time value.

### Legacy Queue Migration And Claim Support

Adds a one-time, atomic legacy withdraw queue migration guarded by `reinitializer(2)`.

Upgrade scripts call the migration selector directly, and protocol-specific legacy checks run through internal hooks instead of external helper functions.

Legacy LP redeem requests remain claimable after the upgrade. Additional safety checks prevent upgrades while unsupported legacy protocol withdrawal queues are still pending.

### Pause / Unpause

Adds a shared ARM pause circuit breaker.

- Operator or owner can pause.
- Only owner/governor can unpause.
- Paused state blocks user-facing flows: swaps, deposits, new LP redeem requests, and LP redeem claims.
- Operational and recovery flows remain available, including allocation, price/buffer updates, fee collection, and base-asset redeem/claim flows.

### Reentrancy Protection

Adds `ReentrancyGuardUpgradeable` protection around liquidity-moving entrypoints, including swap entrypoints, deposits, LP redeem request/claim, and fee collection.

This prevents nested callbacks from reusing stale pre-payout balances during liquidity-moving flows.

### Ethena Adapter Storage And Queue Cleanup

Restores deprecated Ethena ARM storage slots for upgrade layout compatibility.

The Ethena adapter now tracks queued unstaker requests using a total request counter plus FIFO pending index, deriving each unstaker index from request position modulo the unstaker count instead of storing a pending index array.

### Custom Errors and Contract Size Cleanup

Replaces admin/operator-only revert strings with custom errors across ownership, operator, pricing, fee, market, migration, and cap-manager paths.

Custom errors include selector documentation where required.

User-facing revert strings remain largely unchanged for swap, deposit, LP redeem, and adapter flows where preserving existing behavior is useful.

Additional refactors reduce ARM runtime size and keep the ARM implementations below the EIP-170 contract size limit.

## Deployment, Task, and Test Updates

Includes updated deploy/upgrade scripts, ABIs, Hardhat tasks, diagrams, fork tests, smoke tests, invariant tests, and upgrade guard tests for the new architecture.

## Testing

Coverage includes:

- Unit tests for swap liquidity limits, fees, reserves, LP deposits/redeems, pause behavior, reentrancy protection, active-market liquidity sourcing, and adapter behavior
- Fork tests for Lido, EtherFi, Ethena, Origin, and Harvester flows
- Smoke tests for upgraded ARM deployments
- Invariant tests for ARM accounting and redemption queues
- Upgrade guard tests for Lido, EtherFi, and Ethena legacy queue migrations
- Contract size checks confirming the ARM implementations remain under the EIP-170 runtime limit

### Gas comparison

Measured on mainnet fork block `24,846,066`.

| ARM | Scenario | Gas |
| --- | --- | ---: |
| Lido ARM | Current deployed, enough liquidity | 110,041 |
| Lido ARM | Upgraded, enough liquidity | 169,976 |
| Lido ARM | Upgraded, needs Morpho liquidity | 600,216 |
| Ethena ARM | Current deployed, enough liquidity | 93,333 |
| Ethena ARM | Upgraded, enough liquidity | 133,155 |
| Ethena ARM | Upgraded, needs Aave liquidity | 202,789 |

Relative to current deployed:

| ARM | Scenario | Delta |
| --- | --- | ---: |
| Lido ARM | Upgraded, enough liquidity | +59,935 gas / +54.47% |
| Lido ARM | Upgraded, needs Morpho liquidity | +490,175 gas / +445.45% |
| Ethena ARM | Upgraded, enough liquidity | +39,822 gas / +42.67% |
| Ethena ARM | Upgraded, needs Aave liquidity | +109,456 gas / +117.27% |


src/contracts/AbstractARM.sol
clement-ux: What do you think about moving `event` and `struct` into a dedicated interface?
It's purely for aesthetic reasons though, to make the core contract cleaner.

src/contracts/AbstractARM.sol
```diff
@@ -443,19 +492,48 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
+        FeeData memory feeDataMem = feeData;
+        if (feeDataMem.fee == 0) return;
+
+        // Convert the base amount in to liquidity asset terms for fee calculation
+        uint256 convertedAmountIn = _convert(address(inToken), amountIn);
+        if (convertedAmountIn <= amountOut) return;
+
+        // Fee is the percentage of the swap discount
+        feeData.feesAccrued =
+            feeDataMem.feesAccrued + uint128((convertedAmountIn - amountOut) * feeDataMem.fee / FEE_SCALE);
```
clement-ux: Maybe we could use libraries for it? MulDiv and SafeCast?

src/contracts/AbstractARM.sol
```diff
@@ -473,14 +551,24 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
         require(sellT1 >= crossPrice, "ARM: sell price too low");
         require(buyT1 < crossPrice, "ARM: buy price too high");
 
+        // Write state to contract storage
         traderate0 = PRICE_SCALE * PRICE_SCALE / sellT1; // quote (t0) -> base (t1); eg WETH -> stETH
         traderate1 = buyT1; // base (t1) -> quote (t0). eg stETH -> WETH
+        buyLiquidityRemaining = buyAmount;
+        sellLiquidityRemaining = sellAmount;
 
         emit TraderateChanged(traderate0, traderate1);
```
clement-ux: If we have enough space on contract size, it could be worth for integrator/MEV to emit the `buyLiquidityRemaining` and `sellLiquidityRemaining`. WDYT?

src/contracts/AbstractARM.sol
naddison36: I like having the events and structs within the contract.  It think it makes it easier to see the events alongside the functions.

src/contracts/AbstractARM.sol
naddison36: in the past we have avoided emitted events in the swaps to save gas. But gas is becoming less important. We could add a new Swap event that includes the remaining liquidity.

clement-ux: I meant it for `setPrices()` not for swap, but it could make sense too.

src/contracts/AbstractARM.sol
```diff
@@ -1031,24 +1090,38 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
+    function migrateLegacyWithdrawQueue() external onlyOwner {
+        require(withdrawsQueuedShares == 0 && withdrawsClaimedShares == 0, "ARM: already migrated");
+
+        uint256 packedLegacyQueue = reservedWithdrawLiquidity;
+        uint128 legacyQueued = uint128(packedLegacyQueue);
+        uint128 legacyClaimed = uint128(packedLegacyQueue >> 128);
+        require(legacyQueued == legacyClaimed, "ARM: legacy withdrawals pending");
+
+        reservedWithdrawLiquidity = 0;
+    }
```
shahthepro: 🔴  This function is not called as part of the deploy script. But `reservedWithdrawLiquidity` is used as part of `requestRedeem`, `getReserves` and `claimRedeem`, all those methods seem to have no checks to ensure that migration has happened. if there's a pending withdrawal, migration wouldn't work and claim wouldn't work either since it's overriding the slots, am I right?

Here's what Claude has to say when I asked it this question:
```
You're essentially right, but with some nuance worth tracing precisely. Let me walk through what actually happens.                                                                                                                               
                                                                                                                                                                                                                                                   
  The packed legacy slot                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                     
  Pre-upgrade, the slot holds withdrawsQueued | (withdrawsClaimed << 128). Cases:                                                                                                                                                                    
                                                                                                                                                                                                                                                     
  - Fresh ARM, never had a queue: slot = 0. Migration trivially passes.                                                                                                                                                                              
  - ARM with pending requests, none claimed yet: low 128 = queued, high 128 = 0. Not bloated — reservedWithdrawLiquidity reads as just queued. The new code semi-coincidentally treats this as "amount reserved for withdrawals" — semantically    
  similar to the old meaning.                                                                                                                                                                                                                        
  - ARM with claim history: high 128 = cumulative claimed. Bloated. A real Lido ARM with months of claims will sit somewhere in the 10⁵⁰+ range.                                                                                                   
                                                                                                                                                                                                                                                     
  For the deployed Lido / EtherFi / Ethena ARMs that have been operating, you should assume bloated.                                                                                                                                                 
                                                                                                                                                                                                                                                     
  Your two claims                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                   
  "migration wouldn't work" — correct                                                                                                                                                                                                                
                                                                                                                                                                                                                                                   
  migrateLegacyWithdrawQueue requires legacyQueued == legacyClaimed (i.e. the packed slot must be X | X<<128). Any pending pre-upgrade LP request makes legacyQueued > legacyClaimed, and migration reverts.                                         
  
  The upgrade scripts (028 / 029 / 030 / 006) only check the protocol queue (checkNoLegacyLidoWithdrawalRequests checks _deprecatedLidoWithdrawalQueueAmount — the Lido NFT queue, not the LP queue). Nothing in the deploy data calls               
  migrateLegacyWithdrawQueue, and nothing prevents the upgrade going through with LP queue pending.                                                                                                                                                
                                                                                                                                                                                                                                                     
  "claim wouldn't work" — partially right, mechanism is more subtle                                                                                                                                                                                
                                     
  The legacy claim function is gone (old claimRedeem on LidoARM/etc is removed). Legacy entries can only be claimed through the new claimRedeem. Here's what happens for a legacy request post-upgrade:                                              
  
  - The struct grew (added uint128 shares). Legacy entries occupy slots 0–1; slot 2 (shares) reads as 0 — uninitialized.                                                                                                                             
  - claimRedeem line 750: assetsAtClaim = request.shares > 0 ? convertToAssets(request.shares) : request.assets → falls back to request.assets. OK.                                                                                                
  - _burn(address(this), request.shares) → burn 0. OK.                                                                                                                                                                                               
  - request.queued <= claimable() — legacy queued was in liquidity-asset terms, new claimable() is in share terms. For 1:1 share-asset ratios this passes numerically by coincidence; for non-1:1 (e.g. after gains/losses) the comparison is        
  semantically meaningless and could pass or fail wrongly.                                                                                                                                                                                           
  - reservedWithdrawLiquidity -= request.assets — subtracts a small number from the bloated value. Slot is still bloated.                                                                                                                            
                                                                                                                                                                                                                                                     
  So the LP actually gets their funds — claim "works" in the sense that they're paid. The catastrophe is elsewhere.                                                                                                                                  
                                                                                                                                                                                                                                                     
  The actual catastrophic consequence                                                                                                                                                                                                                
                                                                                                                                                                                                                                                   
  reservedWithdrawLiquidity being bloated (~10⁵⁰+) breaks four call sites:                                                                                                                                                                           
  
  1. _ensureLiquidityAvailableForSwap (line 444): requiredLiquidity = amount + reservedWithdrawLiquidity overflows past any plausible liquidityBalance, falls into the market-withdraw path, tries to pull 10⁵⁰ from the market, reverts. All        
  buy-side swaps are dead. Sell-side swaps don't call this, so they still work.                                                                                                                                                                    
  2. _allocate line 1051: currentArmLiquidity = balance - reservedWithdrawLiquidity is hugely negative → _allocate tries to withdraw an absurd amount from the market every time, reverts.                                                           
  3. _requireLiquidityAvailable (line 838): any positive amount fails → collectFees is dead.                                                                                                                                                         
  4. getReserves (line 821): reservedWithdrawLiquidity > liquidityAssets, so returns 0 → off-chain price routers see zero liquidity.                                                                                                                 
                                                                                                                                                                                                                                                     
  And then migration becomes permanently impossible                                                                                                                                                                                                  
                                                                                                                                                                                                                                                     
  migrateLegacyWithdrawQueue requires both withdrawsQueuedShares == 0 && withdrawsClaimedShares == 0. The instant any LP calls requestRedeem post-upgrade, withdrawsQueuedShares becomes non-zero forever. Even after that new request is claimed,   
  withdrawsClaimedShares is also non-zero forever. There is no path to call migrateLegacyWithdrawQueue once any new LP activity has happened.                                                                                                      
                                                                                                                                                                                                                                                     
  So the operational sequence required to land this upgrade safely is:                                                                                                                                                                             
                                     
  1. Drain LP queue (all pending claimed).                                                                                                                                                                                                           
  2. Drain protocol queue (Lido NFT etc).
  3. Freeze new LP activity.                                                                                                                                                                                                                         
  4. Upgrade.                                                                                                                                                                                                                                        
  5. Owner calls migrateLegacyWithdrawQueue immediately, before unfreezing.
  6. Resume.                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                   
  Steps 1, 3, and 5 are not enforced by code, not bundled into upgradeToAndCall, and not asserted by any test. Step 3 isn't even possible — there's no pause mechanism on AbstractARM (no Pausable inheritance, no whenNotPaused modifiers). The     
  owner can't freeze deposits/redeems atomically with the upgrade.                                                                                                                                                                                 
                                                                                                                                                                                                                                                     
  Verdict                                                                                                                                                                                                                                            
                                     
  You're right that this is a footgun. The right fix is one of:                                                                                                                                                                                      
                                                                                                                                                                                                                                                   
  - Bundle the migration into upgradeToAndCall with a multicall (upgradeToAndCallBatch or have the new impl expose a single initializeUpgrade() that both checks the protocol queue and clears the legacy slot in one call). Then step 3 is just     
  "land the upgrade tx".
  - Loosen the migration's pre-condition to "any time the legacy queued == claimed in the slot, even if new activity exists". This lets ops run migration whenever the legacy queue is balanced — though they still need to drain pending pre-upgrade
   requests first.                                                                                                                                                                                                                                   
  - Add a pause modifier to requestRedeem so the owner can freeze new activity between upgrade and migration.
                                                                                                                                                                                                                                                     
  I'm adding this as a new Critical finding (C6) to the review document, with severity equal to C5 — it's effectively the operational version of the same root cause: the migration assumes a clean state that the deploy scripts don't enforce.     
  
```

src/contracts/AbstractARM.sol
```diff
@@ -346,262 +315,408 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
+        BaseAssetConfig storage config = baseAssetConfigs[swapBaseAsset];
+
+        if (!isBuySide) {
+            // Trader sells liquidity asset and buys the base asset.
+            // The ARM buys the liquidity asset and sells the base asset.
+            // ARM prices the base sale at sellPrice.
+            uint256 convertedAmountOut = _convertToAssets(config, amountOut);
+            // amountOut is converted to liquidity terms first, then multiplied by sellPrice
+            // to solve for the required liquidity input.
+            amountIn = convertedAmountOut * config.sellPrice / PRICE_SCALE + 3;
```
shahthepro, line 450: Why add 3 wei? To accounting for rounding issues? Wouldn't 1 wei work in that case?

src/contracts/AbstractARM.sol
```diff
@@ -346,262 +315,408 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
-        // This prevents the ARM making a loss when the base asset is sold at a lower price than it was bought
-        // before the cross price was lowered.
-        if (newCrossPrice < crossPrice) {
-            // Check there is not a significant amount of base assets in the ARM
-            require(IERC20(baseAsset).balanceOf(address(this)) < MIN_TOTAL_SUPPLY, "ARM: too many base assets");
+        require(config.sellPrice >= newCrossPrice, "ARM: sell price too low");
+        require(config.buyPrice < newCrossPrice, "ARM: buy price too high");
+
+        if (newCrossPrice < config.crossPrice) {
+            require(IERC20(priceBaseAsset).balanceOf(address(this)) < MIN_TOTAL_SUPPLY, "ARM: too many base assets");
```
shahthepro: This doesn't account for base asset that's reserved for withdrawal?

src/contracts/adapters/EthenaAssetAdapter.sol
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    function deployUnstakers() external onlyOwner {
+        for (uint256 i = 0; i < MAX_UNSTAKERS; ++i) {
+            if (unstakers[i] == address(0)) unstakers[i] = address(new EthenaUnstaker(address(this), susde));
+        }
+    }
+
+    function setUnstakers(address[MAX_UNSTAKERS] calldata _unstakers) external onlyOwner {
+        unstakers = _unstakers;
```
shahthepro, line 175: Claude flagged this:
```
EthenaAssetAdapter.setUnstakers blindly overwrites — can orphan pending requests

     function setUnstakers(address[MAX_UNSTAKERS] calldata _unstakers) external onlyOwner {
         unstakers = _unstakers;
     }

     If any slot currently holds an unstaker with a non-zero requestShares[unstaker] (still cooling down), replacing it makes the funds locked in the old unstaker's cooldown unrecoverable through this adapter — redeem loops over
     pendingUnstakerIndexes and reads unstakers[idx], which now points to a new (empty) address.

     Recommendation: require requestShares[oldUnstaker] == 0 for every slot being overwritten, OR add a replaceUnstaker(uint8 index, address newUnstaker) that does the check per-slot.
```

src/contracts/AbstractARM.sol
```diff
@@ -6,10 +6,14 @@ import {IERC4626} from "@openzeppelin/contracts/interfaces/IERC4626.sol";
...
 
 /**
  * @title Generic Automated Redemption Manager (ARM)
+ * @notice Coordinates liquidity-provider shares, two-token swaps, active market allocation, and
+ * protocol-specific redemption adapters for one liquidity asset and one or more supported base assets.
+ * @dev Existing ARM proxies depend on the original storage prefix. New multi-base state is appended after
+ * legacy single-base storage so Lido, EtherFi, Ethena, and Origin ARMs can share this implementation.
  * @author Origin Protocol Inc
  */
 abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
```
shahthepro: We usually have a method to recover tokens (`transferToken`), should we have something like that here for non-base assets?

src/contracts/AbstractARM.sol
```diff
@@ -611,292 +726,250 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
         });
 
-        // burn redeemer's shares
-        _burn(msg.sender, shares);
-
-        // Remove the redeemed assets from the last available assets
-        lastAvailableAssets -= SafeCast.toInt128(SafeCast.toInt256(assets));
-
+        // Escrow the redeemer's shares so they stay in totalSupply() and share losses/gains pro-rata.
+        _transfer(msg.sender, address(this), shares);
```
shahthepro, line 810: There seems to be no `shares > 0` check here. This `_transfer` could fail for zero value transfers but having an explicit error would be nice

src/contracts/AbstractARM.sol
```diff
@@ -611,292 +726,250 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
-        // We need to check there is enough liquidity assets to cover the fees being collect from this ARM contract.
-        // We could try the transfer and let it revert if there are not enough assets, but there is no error message with
-        // a failed WETH transfer so we spend the extra gas to check and give a meaningful error message.
         require(fees <= IERC20(liquidityAsset).balanceOf(address(this)), "ARM: insufficient liquidity");
 
+        feesAccrued = 0;
         IERC20(liquidityAsset).transfer(feeCollector, fees);
-
         emit FeeCollected(feeCollector, fees);
     }
```
clement-ux: Maybe we can simplify the logic here:

```solidity
function collectFees() public returns (uint256 fees) {
    fees = feesAccrued;

    // Fees can only be collected from unreserved on-hand liquidity.
    require(
        fees + reservedWithdrawLiquidity <= IERC20(liquidityAsset).balanceOf(address(this)),
        "ARM: insufficient liquidity"
    );

    delete feesAccrued;
    IERC20(liquidityAsset).transfer(feeCollector, fees);
    emit FeeCollected(feeCollector, fees);
}
```

With this, we can remove `_requireLiquidityAvailable()`. 
Before, we had a double check that is thus removed.
```solidity
// In _requireLiquidityAvailable()
require(
    amount + reservedWithdrawLiquidityMem <= IERC20(liquidityAsset).balanceOf(address(this)),
    "ARM: Insufficient liquidity"
);
// ---
// In collectFees()
require(fees <= IERC20(liquidityAsset).balanceOf(address(this)), "ARM: insufficient liquidity");
```

This increase the EtherFi Runtime Margin from 5 to 201.

src/contracts/AbstractARM.sol
clement-ux: I'm joining Shahul on this subject, and I think that someone can block us into the migration process. 
- Check the governance proposal
- Deposit+request
- never claim it
- we are stuck.


Should we be able to claim on behalf of someone and remove this require on `claimRedeem()`? :
```solidity
require(request.withdrawer == msg.sender, "Not requester");
```

src/contracts/AbstractARM.sol
```diff
@@ -283,30 +268,21 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
         require(path.length == 2, "ARM: Invalid path length");
```
clement-ux, line 324: To reduce the code size, we could replace all `require(..., "...")` and `revert("...")` by custom errors. 
According to Codex, this could free up to 2kB!

src/contracts/LidoARM.sol
```diff
@@ -2,36 +2,23 @@
...
 /**
  * @title Lido (stETH) Automated Redemption Manager (ARM)
  * @dev This implementation supports multiple Liquidity Providers (LPs) with single buy and sell prices.
  * It also integrates to a CapManager contract that caps the amount of assets a liquidity provider
  * can deposit and caps the ARM's total assets.
- * A performance fee is also collected on increases in the ARM's total assets.
+ * A fee is accrued on discounted base-asset buy swaps.
  * @author Origin Protocol Inc
  */
 contract LidoARM is Initializable, AbstractARM {
```
clement-ux, line 16: Because all the ARM logic is now in the Adapter, I'm suggesting:
- Rename Adapter into ARM, because this is were the "Redemption Manager" lives now
- Remove `LidoARM`, `EtherfiARM` etc. wrapper around `AbstractARM.sol` and deploy it directly.
- Rename `AbstractARM` into something like `OriginPool` maybe? 

This is high level suggestion, to have a better naming/description of the contract themself.
Not sure if this is a great idea, throwing it there for discussion :)

clement-ux (commented):
First review session focused on high level code architecture and not on security.

clement-ux:
Another idea to decrease the size of the contract could be to move the lending part into an external contract, like you did for the `Adapter`. Or to reduce gas, to have the following logic in the `MorphoMarket.sol`:
- `addMarkets()`
- `removeMarket()`
- `setActiveMarket()`
- `allocate()` and `_allocate()`

It shouldn't increase gas cost for swap, even for large swap that need to withdraw from lending market.

src/contracts/AbstractARM.sol
naddison36: Good point. I've added in commit https://github.com/OriginProtocol/arm-oeth/pull/208/changes/28d4fb654fb674ec7638e33e63bb279aeab765dd

src/contracts/AbstractARM.sol
naddison36: because stETH can round more with larger transfers. I've seen 2 wei. 3 wei was to be safe

src/contracts/AbstractARM.sol
naddison36: Good point. Now we value base assets in the withdrawal queue at the cross price, the check in setPrice should include both base assets in the ARM and withdrawal queue

Fixed in commit 
https://github.com/OriginProtocol/arm-oeth/pull/208/changes/135f1c647b8b23a67a9579da18de8f17ed901639

src/contracts/AbstractARM.sol
naddison36: You are correct. I missed making it so the Operator can claim a redeem on behalf of someone else. Fixed in commit https://github.com/OriginProtocol/arm-oeth/pull/208/changes/1988a988597ec3120f038077b0a22b608d94e7d4

src/contracts/adapters/EthenaAssetAdapter.sol
naddison36: Good one. Fixed in commit https://github.com/OriginProtocol/arm-oeth/pull/208/changes/6ef36ee0e85cabd89e141464afdbaab95be68e62

src/contracts/AbstractARM.sol
naddison36: We are super tight on the contract size. That would definitely push the ARM contracts to be oversized

src/contracts/AbstractARM.sol
naddison36: `_transfer` itself allows zero-value transfers in OZ v5, but we should probably reject zero-share redeems explicitly so users cannot create empty withdrawal requests / advance the queue. The problem is we don't have much room to add this check.

The same goes for zero deposits in PR https://github.com/OriginProtocol/arm-oeth/pull/229

I'll see if I can get the contract size down

src/contracts/LidoARM.sol
naddison36: I think the wrappers help the off-chain code get the correct contract. It also looks better in Etherscan.

I think we keep the ARM name for now. It will eventually go if we move to more flexible investment infrastructure.

I have stripped out more of the legacy variables from the wrappers. That's given us some more contract size space.

src/contracts/AbstractARM.sol
naddison36: I've been holding off doing this change for backward compatibility reasons.

src/contracts/AbstractARM.sol
naddison36: This is a good simplification. Done with commit https://github.com/OriginProtocol/arm-oeth/pull/208/changes/75f5756c7a57422bf3cfd33bd2e5de1bad81fd7c

src/contracts/AbstractARM.sol
shahthepro: It looks good. But I also wonder if we should pause new requests until the migration happened. Otherwise, someone could theoretically create requests constantly to DoS from us upgrading it. Also, what about the storage slot overriding behaviour? If we delay the migration after upgrade, it'll mess up the storage slots

src/contracts/AbstractARM.sol
shahthepro: Got it

shahthepro: A comment about this would be good

src/contracts/adapters/EthenaAssetAdapter.sol
shahthepro: Looks good but I think we probably need to be cautious of the length of the unstaker array. If it's too long to fit in a block tx, we will be blocked from using this function without an upgrade

src/contracts/AbstractARM.sol
shahthepro: Ah, that makes sense

src/contracts/adapters/EthenaAssetAdapter.sol
naddison36: `MAX_UNSTAKERS` is currently set to 42. I think that's small enough

src/contracts/AbstractARM.sol
naddison36: I'm trying to squeeze in pause in PR https://github.com/OriginProtocol/arm-oeth/pull/226

src/contracts/AbstractARM.sol
clement-ux: Indeed, I haven't thought about backward compatibility for this 👍

src/contracts/AbstractARM.sol
shahthepro: Perhaps I'm missing something but I still think this is gonna mess up that storage slot: https://github.com/OriginProtocol/arm-oeth/blob/nicka/withdraw-on-swap/src/contracts/AbstractARM.sol#L756-L757

claimRedeem modifies that variable as well and it's not excluded during pause. This is fine for newly deployed ARM contracts. But for the upgrades, would be better to test this on fork first

shahthepro: Any claimRedeem will also make the migration to not run since it messes up that storage slot

src/contracts/AbstractARM.sol
```diff
@@ -346,262 +321,411 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
+            // The ARM buys the base asset and sells the liquidity asset.
+            // ARM prices the base purchase at buyPrice.
+            uint256 convertedAmountIn = _convertToAssets(config, amountIn);
+            // buyPrice is liquidity assets per base asset. Since convertedAmountIn is the
+            // base input expressed in liquidity terms, multiply by buyPrice to get liquidity output.
+            amountOut = convertedAmountIn * config.buyPrice / PRICE_SCALE;
+
+            _accrueSwapFee(config.buyPrice, config.crossPrice, amountOut);
+            _ensureLiquidityAvailableForSwap(amountOut);
         }
```
clement-ux, line 419: nit: this can be simplified into:
```solidity
uint256 convertedAmountIn = _convertToShares(config, amountIn);
if (!isBuySide) {
    // Trader sells liquidity asset and buys the base asset.
    // The ARM buys the liquidity asset and sells the base asset.
    // ARM prices the base sale at sellPrice.
    // sellPrice is liquidity assets per base asset, so divide liquidity input by sellPrice
    // to get the base output owed to the trader.
    amountOut = convertedAmountIn * PRICE_SCALE / config.sellPrice;
} else {
    // Trader sells base asset and buys the liquidity asset.
    // The ARM buys the base asset and sells the liquidity asset.
    // ARM prices the base purchase at buyPrice.
    // buyPrice is liquidity assets per base asset. Since convertedAmountIn is the
    // base input expressed in liquidity terms, multiply by buyPrice to get liquidity output.
    amountOut = convertedAmountIn * config.buyPrice / PRICE_SCALE;

    _accrueSwapFee(config.buyPrice, config.crossPrice, amountOut);
    _ensureLiquidityAvailableForSwap(amountOut);
}
```

src/contracts/AbstractARM.sol
```diff
@@ -346,262 +321,411 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
-        // The previous assignment assumed token0 is be the liquidity asset.
-        // If not, swap the reserves
-        if (address(token0) == baseAsset) (reserve0, reserve1) = (reserve1, reserve0);
+    /// @dev Accrue fees on discounted buy-side swaps using the recognized NAV gain.
+    /// @param buyPrice Price the ARM paid for the base asset.
+    /// @param crossPrice Price used to value the base asset in totalAssets().
+    /// @param amountOut Liquidity asset amount paid out by the ARM.
+    function _accrueSwapFee(uint256 buyPrice, uint256 crossPrice, uint256 amountOut) internal {
+        uint256 feeMultiplier =
+            buyPrice == 0 ? 0 : (crossPrice - buyPrice) * uint256(fee) * PRICE_SCALE / (buyPrice * FEE_SCALE);
```
clement-ux: I think we already tested it before, should we store this value and update it at every call of `setPrices()` and `setFee()`. Like this we avoid this calculation at every swap and thus reduce gas cost for a swap.

src/contracts/AbstractARM.sol
```diff
@@ -346,262 +321,411 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
-    function getReserves() external view returns (uint256 reserve0, uint256 reserve1) {
-        // The amount of liquidity assets (WETH) that is still to be claimed in the withdrawal queue
-        uint256 outstandingWithdrawals = withdrawsQueued - withdrawsClaimed;
+    /// @dev Consume the per-base liquidity limit for the current swap direction.
+    /// Buy-side limits are denominated in liquidity assets. Sell-side limits are denominated in base assets.
+    /// @param config Base asset config whose liquidity limit is consumed.
+    /// @param isBuySide True when the ARM buys base asset and pays out liquidity asset.
+    /// @param amountOut Amount of output token sent by the ARM.
+    function _consumeSwapLiquidityLimit(BaseAssetConfig storage config, bool isBuySide, uint256 amountOut) internal {
+        uint256 remaining = isBuySide ? config.buyLiquidityRemaining : config.sellLiquidityRemaining;
```
clement-ux: Do we plan having `config.buyLiquidityRemaining` and `config.sellLiquidityRemaining` set to `type(uint256).max`? For example in a situation where we want to disable swap liquidity limits?

If so what do you think about returning earlier, to avoid writing new value?

src/contracts/AbstractARM.sol
```diff
@@ -1031,24 +1083,38 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
         require(_armBuffer <= 1e18, "ARM: invalid arm buffer");
         armBuffer = _armBuffer;
-
         emit ARMBufferUpdated(_armBuffer);
     }
+
+    /// @notice Clear the legacy packed asset queue counter slot during the Model A upgrade.
+    /// @dev The reused slot previously packed `withdrawsQueued` in the low 128 bits and
+    /// `withdrawsClaimed` in the high 128 bits. It may be nonzero even when the old queue
+    /// is fully drained, so upgrade scripts should call this with `upgradeToAndCall`.
```
sparrowDom: nit: comment upgrade scripts don't call this in "upgradeToAndCall"

src/contracts/AbstractARM.sol
sparrowDom: It is annoying to address this issue while also battling with the contract size limits. There is another way this could be done which is the cleanest, but operationally very annoying: 
- upgrade the existing ARMs by adding the ability to pause (we would need to be mindful of deprecating this storage slot in this PR / future contract upgrade) and claim withdrawals on other user's behalf
- once upgrade lands, and withdrawals are claimable claim all of them
- upgrade the arms to this implementation where the withdrawal queues are drained. The migration call happens atomically in the deploy script's "upgradeToAndCall"

src/contracts/AbstractARM.sol
sparrowDom: https://github.com/OriginProtocol/arm-oeth/pull/208/changes/873b46b9b2ca0b3596861f136351ee3e297f5b6d

src/contracts/AbstractARM.sol
clement-ux: Two things here:
- We are on the edge from having an overflow here. If `buyPrice ≈ 0 -> (crossPrice - buyPrice) ≈ 1e36`, maximum for fee is `5e4` , then `*1e36` -> `≈ 5e76`. Maximum before overflow is `≈1.15e77`.
- We are rounding down, 2 times (one in `feeMultiplier`, one in `feesAccrued`), and not in the favor of the protocol.

2 solutions:
- Use Math.mulDiv
- Change `PRICE_SCALE` by `1e18`.

I'm leaning toward option 1 (even though it is impossible to have an overflow with current code), because it rounds in favor of the protocol :
```solidity
function _accrueSwapFee(uint256 buyPrice, uint256 crossPrice, uint256 amountOut) internal {
    if (buyPrice == 0) return;

    uint256 feeAmount = Math.mulDiv(
        amountOut,
        (crossPrice - buyPrice) * uint256(fee),
        buyPrice * FEE_SCALE,
        Math.Rounding.Ceil
    );

    feesAccrued = SafeCast.toUint128(feesAccrued + feeAmount);
}
```

src/contracts/AbstractARM.sol
sparrowDom: this one should be non-reentrant

sparrowDom: this one should be non-reentrant

src/contracts/AbstractARM.sol
```diff
@@ -346,262 +321,411 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
+            // Trader sells base asset and buys the liquidity asset.
+            // The ARM buys the base asset and sells the liquidity asset.
+            // ARM prices the base purchase at buyPrice.
+            uint256 convertedAmountIn = _convertToAssets(config, amountIn);
+            // buyPrice is liquidity assets per base asset. Since convertedAmountIn is the
+            // base input expressed in liquidity terms, multiply by buyPrice to get liquidity output.
+            amountOut = convertedAmountIn * config.buyPrice / PRICE_SCALE;
+
+            _accrueSwapFee(config.buyPrice, config.crossPrice, amountOut);
+            _ensureLiquidityAvailableForSwap(amountOut);
```
sparrowDom: 🟠  `_ensureLiquidityAvailableForSwap` can do an external 4626 call and `_consumeSwapLiquidityLimit` only later on changes the storage state. A maicious 4626 market could re-enter this function and cause damage. 

Maybe move `_consumeSwapLiquidityLimit(config, isBuySide, amountOut);` into the if statement: 
```
 if (!isBuySide) {
    // Trader sells liquidity asset and buys the base asset.
    // The ARM buys the liquidity asset and sells the base asset.
    // ARM prices the base sale at sellPrice.
    uint256 convertedAmountIn = _convertToShares(config, amountIn);
    // sellPrice is liquidity assets per base asset, so divide liquidity input by sellPrice
    // to get the base output owed to the trader.
    amountOut = convertedAmountIn * PRICE_SCALE / config.sellPrice;
   _consumeSwapLiquidityLimit(config, isBuySide, amountOut);
} else {
    revert("ARM: Invalid in token");
    // Trader sells base asset and buys the liquidity asset.
    // The ARM buys the base asset and sells the liquidity asset.
    // ARM prices the base purchase at buyPrice.
    uint256 convertedAmountIn = _convertToAssets(config, amountIn);
    // buyPrice is liquidity assets per base asset. Since convertedAmountIn is the
    // base input expressed in liquidity terms, multiply by buyPrice to get liquidity output.
    amountOut = convertedAmountIn * config.buyPrice / PRICE_SCALE;

    _accrueSwapFee(config.buyPrice, config.crossPrice, amountOut);
    _consumeSwapLiquidityLimit(config, isBuySide, amountOut);
    _ensureLiquidityAvailableForSwap(amountOut);
}
```

src/contracts/AbstractARM.sol
```diff
@@ -346,262 +321,411 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
+        uint256 buyAmount,
+        uint256 sellAmount
+    ) external onlyOperatorOrOwner {
+        BaseAssetConfig storage config = baseAssetConfigs[priceBaseAsset];
+        require(config.adapter != address(0), "ARM: unsupported asset");
+        require(sellPrice >= config.crossPrice, "ARM: sell price too low");
+        require(buyPrice < config.crossPrice, "ARM: buy price too high");
+
+        config.buyPrice = SafeCast.toUint128(buyPrice);
+        config.sellPrice = SafeCast.toUint128(sellPrice);
```
clement-ux, line 640: I really like having prices like this, and not having `traderade0` "inverted"`PRICE_SCALE * PRICE_SCALE / sellT1;` like before 👍

src/contracts/AbstractARM.sol
clement-ux: If `isBuySide == false`, we don’t check whether there is enough liquidity before proceeding. As a result, the transaction does not revert with a clear error message, but instead fails with a `panic: arithmetic-underflow/overflow`.

This is not a critical issue since this path is not the most commonly used one, but should we add an explicit liquidity check so it reverts with a proper error message?

src/contracts/AbstractARM.sol
```diff
@@ -346,262 +321,411 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
+    /// @param isBuySide True when the ARM buys base asset and pays out liquidity asset.
+    /// @param amountOut Amount of output token sent by the ARM.
+    function _consumeSwapLiquidityLimit(BaseAssetConfig storage config, bool isBuySide, uint256 amountOut) internal {
+        uint256 remaining = isBuySide ? config.buyLiquidityRemaining : config.sellLiquidityRemaining;
+        require(amountOut <= remaining, "ARM: Insufficient liquidity");
+
+        remaining -= amountOut;
+        if (isBuySide) config.buyLiquidityRemaining = SafeCast.toUint128(remaining);
+        else config.sellLiquidityRemaining = SafeCast.toUint128(remaining);
+    }
```
clement-ux: Maybe we could refactor it in way that doesn't mix `?` and `if/else`, and that consume less gas:
```solidity
function _consumeSwapLiquidityLimit(BaseAssetConfig storage config, bool isBuySide, uint256 amountOut) internal {
    if (isBuySide) {
        uint256 remaining = config.buyLiquidityRemaining;
        require(amountOut <= remaining, "ARM: Insufficient liquidity");

        unchecked {
            config.buyLiquidityRemaining = uint128(remaining - amountOut);
        }
    } else {
        uint256 remaining = config.sellLiquidityRemaining;
        require(amountOut <= remaining, "ARM: Insufficient liquidity");

        unchecked {
            config.sellLiquidityRemaining = uint128(remaining - amountOut);
        }
    }
}
```
Using unchecked saves:
  - stETH -> WETH: 126 gas, ~0.18% gas
  - WETH  -> stETH: 130 gas, ~0.21% gas

But if we don't want to use unchecked, we still save:
  - stETH -> WETH: 84 gas, ~0.11% gas
  - WETH  -> stETH: 68 gas, ~0.14% gas

Fun fact, doing `uint256 remaining = config.buyLiquidityRemaining;` this only save 2-4 gas 😅, so we can even think about:
```solidity
```solidity
function _consumeSwapLiquidityLimit(BaseAssetConfig storage config, bool isBuySide, uint256 amountOut) internal {
    if (isBuySide) {
        require(amountOut <= config.buyLiquidityRemaining, "ARM: Insufficient liquidity");
        unchecked {
            config.buyLiquidityRemaining -=  uint128(amountOut);
        }
    } else {
        require(amountOut <= config.sellLiquidityRemaining, "ARM: Insufficient liquidity");
        unchecked {
            config.sellLiquidityRemaining -= uint128(amountOut);
        }
    }
}
```
```

File: src/contracts/AbstractARM.sol [hunk 19]
```diff
@@ -346,262 +321,412 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
+        returns (address swapBaseAsset, bool isBuySide)
+    {
+        if (inToken == liquidityAsset && baseAssetConfigs[outToken].adapter != address(0)) {
+            return (outToken, false);
+        }
+        if (outToken == liquidityAsset && baseAssetConfigs[inToken].adapter != address(0)) {
+            return (inToken, true);
+        }
+        revert("ARM: Invalid swap assets");
+    }
```
clement-ux: I think we can simplify this into:
```solidity
/// @dev Resolve the supported base asset config from a 2-token swap pair.
/// @param inToken Swap input token address.
/// @param outToken Swap output token address.
/// @return config Supported base asset config involved in the swap.
/// @return isBuySide True when the ARM buys base asset and pays out liquidity asset.
function _getSwapBaseAssetConfig(address inToken, address outToken)
    internal
    view
    returns (BaseAssetConfig storage config, bool isBuySide)
{
    address liquidityAssetCache = liquidityAsset;

    if (outToken == liquidityAssetCache) {
        config = baseAssetConfigs[inToken];
        if (config.adapter != address(0)) return (config, true);
    } else if (inToken == liquidityAssetCache) {
        config = baseAssetConfigs[outToken];
        if (config.adapter != address(0)) return (config, false);
    }

    revert("ARM: Invalid swap assets");
}
```
And change in `_swapExactTokensForTokens()`:
```diff
- (address swapBaseAsset, bool isBuySide) = _getSwapBaseAsset(address(inToken), address(outToken));
- BaseAssetConfig storage config = baseAssetConfigs[swapBaseAsset];
+ (BaseAssetConfig storage config, bool isBuySide) = _getSwapBaseAssetConfig(address(inToken), address(outToken));
```
- It avoids doing the same baseAssetConfigs[...] lookup twice.
- It makes the helper return exactly what callers need: the config and swap direction.
- It saves a small amount of gas on every valid swap path (approx -0.2% gas).

src/contracts/AbstractARM.sol
```diff
@@ -346,262 +321,412 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
-        // The amount of liquidity assets (WETH) that is still to be claimed in the withdrawal queue
-        uint256 outstandingWithdrawals = withdrawsQueued - withdrawsClaimed;
+    /// @dev Consume the per-base liquidity limit for the current swap direction.
+    /// Buy-side limits are denominated in liquidity assets. Sell-side limits are denominated in base assets.
+    /// @param config Base asset config whose liquidity limit is consumed.
+    /// @param isBuySide True when the ARM buys base asset and pays out liquidity asset.
+    /// @param amountOut Amount of output token sent by the ARM.
+    function _consumeSwapLiquidityLimit(BaseAssetConfig storage config, bool isBuySide, uint256 amountOut) internal {
+        uint256 remaining = isBuySide ? config.buyLiquidityRemaining : config.sellLiquidityRemaining;
+        require(amountOut <= remaining, "ARM: Insufficient liquidity");
```
clement-ux: We are using the same require message as the one used in `_ensureLiquidityAvailableForSwap()`, should we have a different message? I'm thinking about that because the integrator might want to know if they cannot swap this amount because there is not enough token in ARM + lending market, or if the limit for that price has been reached. Not sure about that though.

src/contracts/AbstractARM.sol
naddison36: I've moved the migration into `migrateLegacyWithdrawQueue` which can only be called once by the Owner. This then calls can abstract `_checkNoLegacyWithdrawQueue` to check if there are any outstanding withdrawal requests.
claimRedeem can be called by the Operator so that can clear any old redeem requests.
Yes, someone could still brick the migration by continuing to request redeems before the Operator claims them. Ideally, the ARM would be pausable but the contract size is too big at the moment.
Changes are in commit https://github.com/OriginProtocol/arm-oeth/pull/208/changes/4013e6170b25692c4de18f6fa5c232ca84326d12

src/contracts/AbstractARM.sol
naddison36: We did that max liquidity limits but I removed it to save contract size and swap gas costs. The limit can be set high enough that there is effectively no liquidity limit on a swap price.

src/contracts/AbstractARM.sol
naddison36: This is now accurate. The Ethena, EtherFi, and Lido upgrade scripts now all call `migrateLegacyWithdrawQueue()` as the `upgradeToAndCall` payload.
Fixed in commit https://github.com/OriginProtocol/arm-oeth/pull/208/changes/4013e6170b25692c4de18f6fa5c232ca84326d12

src/contracts/AbstractARM.sol
naddison36: we have never had enough contract space to add non-reentrant. The latest failed attempt in in PR https://github.com/OriginProtocol/arm-oeth/pull/227

src/contracts/AbstractARM.sol
naddison36: I originally had `_consumeSwapLiquidityLimit` in the if statements but moved it out as it reduced the amount of code. But given we currently don't have non-reentrancy protection on swaps, I've moved it back into the if statements.
Changed in commit https://github.com/OriginProtocol/arm-oeth/pull/208/changes/ead4681e226db976cf8a88ea50537ac5f437a31d

naddison36: Ideally, yes, we would have a clear error message. The problem is the ARM contracts are already at their size limit. I'll see if I can make some room

naddison36: I managed to fit the base liquidity check in by refactoring the swap helper functions.
See commit https://github.com/OriginProtocol/arm-oeth/pull/208/changes/a5218d081f47666870720115deebb2773b2538f3

src/contracts/AbstractARM.sol
naddison36: `_convertToAssets` needs to be used when the ARM is buying base assets, not `_convertToShares`

src/contracts/AbstractARM.sol
naddison36: `Math.mulDiv` pushes the contract size over the limit. I've put a min buyPrice limit in 
`require(buyPrice >= MAX_CROSS_PRICE_DEVIATION && buyPrice < config.crossPrice, "ARM: invalid buy price");`

See commit https://github.com/OriginProtocol/arm-oeth/pull/208/changes/523736e2b44674e402dac3835d7ef740f054f06f

src/contracts/AbstractARM.sol
naddison36: Changes are in commit https://github.com/OriginProtocol/arm-oeth/pull/208/changes/88a33675128139f6fefe25eb1b7b71a2cce12f0f

src/contracts/AbstractARM.sol
clement-ux: Indeed, by mad on this.

src/contracts/AbstractARM.sol
clement-ux: Yes, although there’s still a small rounding issue. It’s not critical, but it would be nice to make it as accurate as possible.

src/contracts/AbstractARM.sol
naddison36: Good idea. The implementation is a little different. see commit https://github.com/OriginProtocol/arm-oeth/pull/208/changes/97f5b58aa593d0dc63f50166a68da6264421cb40

src/contracts/AbstractARM.sol
naddison36: If we had spare contract space then I'd agree and add a separate error message. I'll leave for now while the contract space is tight

src/contracts/AbstractARM.sol
```diff
@@ -346,262 +321,423 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
+    ) private {
+        uint256 remaining;
+        if (isBuySide) {
+            _accrueSwapFee(config.buyPrice, config.crossPrice, amountOut);
+            remaining = config.buyLiquidityRemaining;
+            require(amountOut <= remaining, "ARM: Insufficient liquidity");
+            unchecked {
+                config.buyLiquidityRemaining = uint128(remaining - amountOut);
+            }
+            _ensureLiquidityAvailableForSwap(amountOut);
```
clement-ux, line 526: There is a small asymmetry here:  

`If` side:
1. Check against buyLiquidity
2. Ensure there is enough tokens

`Else` side:
1. Ensure there is enough tokens
2. Check against buyLiquidity

src/contracts/AbstractARM.sol
sparrowDom: great I really like how you added the re-entrency test for this. 
Should we also include a comment explaining why this is in the if statement? Could prevent future "contract size optimisations"

src/contracts/AbstractARM.sol
naddison36: Fixed with commit https://github.com/OriginProtocol/arm-oeth/pull/208/changes/59504bdee9c7d3554b0778526dd13290a28e7164

naddison36: The reentrancy PR has been merged with commit https://github.com/OriginProtocol/arm-oeth/pull/208/changes/59504bdee9c7d3554b0778526dd13290a28e7164

src/contracts/AbstractARM.sol
```diff
@@ -260,348 +299,487 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
+    /// @param assets Liquidity assets to deposit.
+    /// @param receiver Account that receives minted LP shares.
+    /// @return shares LP shares minted.
     function _deposit(uint256 assets, address receiver) internal returns (uint256 shares) {
-        // Do not allow deposits if the ARM can not meet all its withdrawal obligations.
-        require(totalAssets() > MIN_TOTAL_SUPPLY || withdrawsQueued == withdrawsClaimed, "ARM: insolvent");
-
-        // Calculate the amount of shares to mint after the performance fees have been accrued
-        // which reduces the available assets, and before new assets are deposited.
+        require(totalAssets() > MIN_TOTAL_SUPPLY || reservedWithdrawLiquidity == 0, "ARM: insolvent");
```
clement-ux: I'm wondering if we are really protecting from depositing when ARM is insolvent. 
The attacker can still give 0.01 ETH to ARM (donation), which unblock this require, then deposit and get way more shares that expected. 
I agree that this is still better than returning 0 though. But now that we have the pausing mechanism, I'm wondering if this is still relevant. WDYT?

src/contracts/AbstractARM.sol
```diff
@@ -611,334 +789,272 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
 
-        // In the scenario where the ARM has made a loss from after the redeem request, the asset value of
+        // In the scenario where the ARM has made a loss after the redeem request, the asset value of
         // the redeemed shares at the time of the claim is used.
-        // This can happen if there was a significant slashing event on the base asset, eg stETH, after the redeem request was made.
+        // This can happen if there was a significant slashing event on the base asset, eg stETH,
+        // after the redeem request was made.
         uint256 assetsAtClaim = request.shares > 0 ? convertToAssets(request.shares) : request.assets;
         // Use the minimum of the asset value of the redeemed shares at request or claim.
         assets = request.assets < assetsAtClaim ? request.assets : assetsAtClaim;
```
clement-ux: I'm not sure, but I think this line isn't necessary. 
If `request.shares == 0` then `convertToAssets(request.shares)` will return 0 too. And to have `request.shares == 0` this implies that user call `requestRedeem(0)`. Calling `requestRedeem(0)` force `request.asset == 0` too.

So we can maybe have:
```diff
- uint256 assetsAtClaim = request.shares > 0 ? convertToAssets(request.shares) : request.assets;
- assets = request.assets < assetsAtClaim ? request.assets : assetsAtClaim;
+ assets = min(request.assets, convertToAssets(request.shares));
```

I agree with Shahul that we should not allow `requestRedeem(0)` though.

sparrowDom (approved):
LGTM

PR #212, opened 2026-04-13
WIP ARM fees charged on buying base assets

naddison36:
## Changes

- Replace the old performance-fee-on-asset-growth model with swap-only fee accrual, so fees accrue only on discounted base-asset buy swaps.
- Update swap fee math to charge `(convertedBaseAmountIn - liquidityAmountOut) * fee / FEE_SCALE` on qualifying swaps, including appreciating-base-asset paths like `sUSDe -> USDe`.
- Change `totalAssets()` to subtract only stored accrued swap fees instead of fees derived from passive asset growth.
- Replace legacy `lastAvailableAssets` fee accounting with explicit stored `feesAccrued`, and stop maintaining the deprecated legacy field after upgrade.
- Reuse the old fee storage slot via a packed `FeeData` struct that holds `fee` and `feesAccrued` in a single slot.
- Keep `feesAccrued` at a `1 wei` sentinel after collection to avoid a more expensive `0 -> nonzero` storage write on the next fee-bearing swap.
- Add migration initializers to LidoARM, EtherFiARM, EthenaARM, and OriginARM to clear the reused legacy fee storage during upgrade.
- Preserve the external `fee()` and `feesAccrued()` interfaces while moving internal fee state to the packed struct layout.

## Gas impact

- Before upgrade, the full Lido `stETH -> WETH` fee-bearing swap transaction used `96,864` gas.  
- After upgrade, the full swap transaction used `105,354` gas.  
- Net change: `+8,490` gas, which is about an `8.8%` increase.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -30,6 +30,8 @@ make coverage         # LCOV coverage report
 
 Linting: `forge fmt --check` (Solidity), `pnpm lint` (JS), `pnpm prettier:check` (JS)
 
+When changing Solidity files, run `forge fmt` before finishing your work.
+
 ## Environment Setup
 
 Copy `.env.example` to `.env` and set `PROVIDER_URL` (Ethereum RPC) and `SONIC_URL` (Sonic RPC). Fork tests require these RPC endpoints.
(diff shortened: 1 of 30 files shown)


naddison36:
Closing as this is now in https://github.com/OriginProtocol/arm-oeth/pull/208

PR #213, opened 2026-04-21
WIP Curator changes

naddison36:
## Changes

- Curator can control the allocation to the ending market
	- drop the current ARM buffer
- Curator can limit the amount that can be swapped for the buy and sell prices
	- The amount reduces with each swap. No more swaps once the amount goes to zero


naddison36:
Closing as this was an experimental branch

PR #214, opened 2026-04-25
Fix Ethena valuation of Aave lending market

naddison36:
The Ethena ARM experienced negative APY on 23 April from block 24944747.

### Root Cause

The ARM’s valuation is done in `totalAssets()`. Inside that, the active lending market is valued via `previewRedeem(allShares)`, so the ARM’s valuation depends on how the underlying ERC-4626 market implements `previewRedeem`.

For Morpho vaults, `previewRedeem` follows the same pricing-conversion path as `convertToAssets`, so it does not directly reflect liquidity constraints. If liquidity is insufficient, `redeem` reverts, and liquidity awareness shows up in `maxWithdraw` / `maxRedeem` instead.

For the Ethena Aave vault, `previewRedeem` is liquidity-aware and can return less than `convertToAssets` when liquidity is constrained.

Unfortunately, both implementations are compliant with the ERC-4626 specification which is very loose on how exactly vault should be implemented.

### What’s the fix

Change `_availableAssets` used by `totalAssets` to use `convertToAssets` instead of `previewRedeem` when valuing the assets in the lending market. This will make the valuation consistent between Morpho and Aave markets.

### Do the other ARMs have the same problem?

No, the Lido and EtherFi ARMs that use Morpho do not have the same problem as the Ethena ARM.

For the Morpho V1.1 and V2 Vaults, `previewRedeem` returns the same as `convertToAssets`  which is a pricing conversion. `redeem` will revert if there is not enough liquidity.

For Morpho V1 Vault, `maxRedeem` is liquidity aware and will return the max allowed given the current liquidity. For Morpho V2 Vault, `maxRedeem` returns zero, which is not ERC-4626 compliant.

image


clement-ux (approved):
This change is indeed important and necessary. 
The new logic is better and can be used across all ERC4626 vault, regardless of whether they are liquidity-aware.

PR #217, opened 2026-05-04
Using swap multiplier to calculate accrued fee

naddison36:
Measured with mainnet fork block 24,846,066, deposit 1,000 WETH, then ARM buys stETH for exactly 100 WETH out.

| Target | Branch / commit | Gas used |
|---|---|---:|
| Currently deployed Lido ARM | mainnet proxy impl | `110,041` |
| Feature branch | `nicka/withdraw-on-swap` @ `2b53b88` | `121,147` |
| Current branch | `nicka/swap-fee-simple` @ `2f69f62` | `117,875` |


src/contracts/AbstractARM.sol
```diff
@@ -569,6 +560,7 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
         traderate1 = buyT1; // base (t1) -> quote (t0). eg stETH -> WETH
         buyLiquidityRemaining = buyAmount;
         sellLiquidityRemaining = sellAmount;
+        _updateSwapFeeMultiplier(buyT1, fee);
```
clement-ux, line 575: nit: It feels weird to me to have `buyT1` and `feeMem` as args.
Of course it reduces gas:
```
With args:    63590
Without args: 63684 (+0.14%)
```
But imo it's more weird that it save gas.

src/contracts/AbstractARM.sol
```diff
@@ -494,28 +488,25 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
         // Return if not buying the base asset
         if (address(inToken) != baseAsset || address(outToken) != liquidityAsset) return;
 
-        FeeData memory feeDataMem = feeData;
-        if (feeDataMem.fee == 0) return;
-
-        // Convert the base amount in to liquidity asset terms for fee calculation
-        uint256 convertedAmountIn = _convert(address(inToken), amountIn);
-        if (convertedAmountIn <= amountOut) return;
+        if (swapFeeMultiplier == 0) return;
```
clement-ux: `swapFeeMultiplier==0` has very low chances happening, should we remove this check to slightly decrease gas?

src/contracts/AbstractARM.sol
```diff
@@ -133,8 +130,12 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
     uint256 public buyLiquidityRemaining;
     /// @notice Remaining liquidity-asset amount that can be sold into the ARM at the current sell price.
     uint256 public sellLiquidityRemaining;
+    /// @notice Multiplier used to accrue liquidity-asset fees from discounted base-asset buy swaps.
+    uint128 public swapFeeMultiplier;
+    /// @notice Accrued swap fees denominated in the liquidity asset.
+    uint128 public feesAccrued;
 
-    uint256[36] private _gap;
+    uint256[34] private _gap;
```
clement-ux: I think it should be `uint256[35] private _gap;`, because `swapFeeMultiplier` and `feesAccrued` will fit into a single slot.

src/contracts/AbstractARM.sol
```diff
@@ -494,28 +488,25 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
      * No fee is accrued when the ARM sells the base asset or when the executed swap is at or above par.
      * @param inToken The token sent into the ARM by the swapper.
      * @param outToken The token sent out from the ARM to the swapper.
-     * @param amountIn The amount of base asset received by the ARM.
      * @param amountOut The amount of liquidity asset paid out by the ARM.
      */
-    function _accrueSwapFee(IERC20 inToken, IERC20 outToken, uint256 amountIn, uint256 amountOut) internal {
+    function _accrueSwapFee(IERC20 inToken, IERC20 outToken, uint256 amountOut) internal {
         // Return if not buying the base asset
         if (address(inToken) != baseAsset || address(outToken) != liquidityAsset) return;
```
clement-ux: Maybe we can save a bit more gas here by removing `address(outToken) != liquidityAsset` as it should be already checked in `_swapExact()`

src/contracts/AbstractARM.sol
```diff
@@ -909,11 +901,21 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
 
         emit FeeUpdated(_fee);
     }
 
+    /// @dev Updates the fee multiplier used to accrue liquidity-asset fees on discounted base-asset buy swaps.
+    /// @param buyT1 The price the ARM buys base assets from traders, scaled to 36 decimals.
+    /// @param feeMem The swap fee measured in basis points.
+    function _updateSwapFeeMultiplier(uint256 buyT1, uint256 feeMem) internal {
+        swapFeeMultiplier = SafeCast.toUint128(
+            buyT1 == 0 || feeMem == 0 ? 0 : (PRICE_SCALE - buyT1) * feeMem * PRICE_SCALE / (buyT1 * FEE_SCALE)
```
clement-ux: There is a small chance that `swapFeeMultiplier == 0` even when `buyT1 != 0` and `fee != 0`: 
- if `fee == 0.01%` `PRICE_SCALE - buyT1 < 10_000`.
- if `fee == 20%` `PRICE_SCALE - buyT1 < 5`   

But this scenario never happened in the past and even if it happens there are no bad consequence.

src/contracts/AbstractARM.sol
naddison36: Codex didn't have args but it means the values had to be re-read from storage when setting the price. That's why I ask it to use params

src/contracts/AbstractARM.sol
naddison36: Good idea. Done in commit https://github.com/OriginProtocol/arm-oeth/pull/217/changes/a77d8d60c7c87ab83685080b2be948a302986ec9

src/contracts/AbstractARM.sol
naddison36: You are correct. Fixed in https://github.com/OriginProtocol/arm-oeth/pull/217/changes/00fe440f4d889bae4e7fea897fda94d6b40fa9ff

src/contracts/AbstractARM.sol
naddison36: Good idea. Done in commit https://github.com/OriginProtocol/arm-oeth/pull/217/changes/85345d346149211a697e5e1ce4acf703c602cdc5

src/contracts/AbstractARM.sol
naddison36: I've removed `|| feeMem` as the multiplier will still be set to zero if `feeMem` is zero. See commit https://github.com/OriginProtocol/arm-oeth/pull/217/changes/ce386523a72def8afda6d8d8ff18022609620107

PR #219, opened 2026-05-05
Reference: Model A loss-sharing for ARM withdrawal queue

clement-ux:
## Context

A recent audit finding ([[OZ AI audit](https://github.com/33e84d46f53c80b49635fb2609928432?pvs=25#34584d46f53c808a83acf270cb3ed2b4)](/33e84d46f53c80b49635fb2609928432?pvs=25#34584d46f53c808a83acf270cb3ed2b4)) shows that the ARM withdrawal queue can become permanently unclaimable after a loss, because the FIFO gate is denominated in pre-loss assets while the payout is reduced by min(request.assets, convertToAssets(request.shares)). Before patching the bug, we need to step back and decide which loss-socialization model the ARM should follow. There are exactly two coherent options.

## Model A — Pro-rata loss sharing

- The redeemer's shares are escrowed (not burned) between request and claim. They remain part of `totalSupply()`.
- Any loss on ARM assets is absorbed proportionally by every share holder, queued or not.
- Any gain accrued between request and claim is forfeited to the remaining LPs at claim time.
- The payout at claim time is `min(request.assets, convertToAssets(shares))`: capped at the request-time snapshot, reduced if a loss happened.
- No "first to request" advantage. No bank-run incentive on bad news. No free-timing-option for the redeemer (the cap removes any reason to delay).
- Trade-off: the redeemer carries downside but keeps no upside during the wait.

## Model B — Senior debt queue

- `requestRedeem` locks in a fixed asset amount (`request.assets`).
- That amount is paid in full at claim time, regardless of subsequent losses.
- Losses are absorbed first by LPs still holding shares, then (if the loss is large enough) by the *last* requesters in the FIFO queue.
- Simple semantics for users and integrators: "you will receive X."
- Trade-off: any rumor of loss triggers a race to `requestRedeem` (bank run). Latecomers eat the loss in full.

The current ARM withdrawal queue uses a hybrid loss-handling model that mixes Model B accounting (burn at request, lock in `request.assets`) with a Model A-style payout reduction (`min(request.assets, convertToAssets(request.shares))`). This combination is the root cause of the audit finding "Withdrawal queue can become unclaimable after losses": the FIFO gate is denominated in pre-loss assets while the payout has been reduced, and `withdrawsClaimed` advances by the pre-loss amount — leaving a phantom liquidity gap that can permanently deadlock the queue.

Full discussion: https://www.notion.so/originprotocol/ARM-How-to-handle-losses-35784d46f53c80e18d70d78cab6c1d52

## What this PR is

This PR is **a reference implementation of Model A** (pro-rata loss sharing via escrow + claim-time pricing). It is meant to make the alternative concrete so the team can compare it side by side with Model B before deciding.

**Model B is the recommended fix** (see the doc). This PR is not the proposed merge path — it exists to ground the discussion.

## Behavioral changes

- Losses (and gains) between request and claim are absorbed pro-rata across every share holder, queued or not.
- No bank-run incentive on bad news: requesting redeem early no longer freezes a payout amount.
- The redeemer no longer sees a fixed `request.assets` — that field becomes a request-time estimate only.
- The FIFO deadlock from the audit finding disappears: the gate and the payout are denominated in the same unit at the same price.

## Code changes (single file: `src/contracts/AbstractARM.sol`)

- **Storage semantics.** `withdrawsQueued`, `withdrawsClaimed`, and `WithdrawalRequest.queued` are re-denominated from assets to **shares**. Storage layout preserved; meaning changed. ⚠ Upgrading a live deployment requires draining the queue first (non-zero pending values would be reinterpreted incorrectly).
- **`requestRedeem`** escrows shares via `_transfer(msg.sender, address(this), shares)` instead of `_burn`. Shares stay in `totalSupply()` so the share price reflects all outstanding claims.
- **`claimRedeem`** pays `convertToAssets(request.shares)` at claim time, burns the escrowed shares right after, and increments `withdrawsClaimed` by `request.shares`.
- **FIFO gate** is now `convertToAssets(pendingShares) <= claimable()`. Same unit, same price, no phantom liquidity gap.
- **`claimable()`** becomes a liquidity snapshot: `balance + maxWithdraw(activeMarket)`. The cumulative `withdrawsClaimed` term is gone.
- **`_availableAssets()`** returns gross assets (no subtraction for outstanding withdrawals). Outstanding shares are now part of `totalSupply()`, so the share price already accounts for them.
- **Liquidity reservation** is dynamic: `_requireLiquidityAvailable`, `_ensureLiquidityAvailableForSwap`, `getReserves`, `_allocate` compute reserved liquidity as `convertToAssets(outstandingShares)` at the current share price.

## Trade-off vs Model B

The main reason **not** to merge this PR is gas:

- `_ensureLiquidityAvailableForSwap` now calls `convertToAssets` → `totalAssets()` → `_availableAssets()` → on a Morpho-backed ARM, `IERC4626(morpho).convertToAssets(allShares)`.
- Morpho's `convertToAssets` is expensive (full market state read, interest accrual, several storage slots).
- Every buy-side swap with a non-empty queue pays this cost. The prior path was zero external calls beyond `balanceOf`.

For an ARM whose value proposition is cheap swaps, this is a meaningful recurring overhead.

## Tests

Tests are intentionally not updated. Many will fail because:

- `withdrawsQueued`/`withdrawsClaimed`/`request.queued` change unit (asset → share)
- The `RedeemRequested` event's `queued` field changes unit
- Payout behavior changes (pro-rata at claim time instead of `min`)

If we ever take this direction, tests would be updated in a follow-up.

## Status

**Reference / discussion. Do not merge.** The proposed fix for the audit finding is the minimal Model B change (drop the `min(...)` in `claimRedeem` so the FIFO gate and the payout become consistent again).

### DIFF
--- a/src/contracts/AbstractARM.sol
+++ b/src/contracts/AbstractARM.sol
@@ -138,7 +138,17 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
     /// @notice Accrued swap fees denominated in the liquidity asset.
     uint128 public feesAccrued;
 
-    uint256[35] private _gap;
+    /// @notice Cumulative `request.assets` ever queued in this ARM's withdrawal queue.
+    /// @dev Asset-denominated mirror of `withdrawsQueued`, used for the swap-side liquidity reservation.
+    /// `withdrawsQueuedAssets - withdrawsClaimedAssets` is an upper bound on the WETH the queue can still
+    /// withdraw (each payout is min(request.assets, convertToAssets(shares)) ≤ request.assets), which lets
+    /// the swap path avoid an expensive convertToAssets / Morpho call.
+    uint128 public withdrawsQueuedAssets;
+    /// @notice Cumulative `request.assets` ever claimed (each claim increments by request.assets, NOT by
+    /// the actual reduced payout, so the difference with `withdrawsQueuedAssets` stays a true upper bound).
+    uint128 public withdrawsClaimedAssets;
+
+    uint256[34] private _gap;
 
     ////////////////////////////////////////////////////
     ///                 Events
@@ -372,9 +382,10 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
     /// @param amount The amount of liquidity assets being sent out of the ARM.
     function _ensureLiquidityAvailableForSwap(uint256 amount) internal {
         uint256 liquidityBalance = IERC20(liquidityAsset).balanceOf(address(this));
-        // Reserve liquidity for queued redemptions, valued at current (loss-adjusted) share price.
-        uint256 outstandingShares = withdrawsQueued - withdrawsClaimed;
-        uint256 reservedAssets = outstandingShares == 0 ? 0 : convertToAssets(outstandingShares);
+        // Upper bound on what the queue can still withdraw. Cheap (no convertToAssets / Morpho call):
+        // each payout is min(request.assets, convertToAssets(shares)) ≤ request.assets, so the cumulative
+        // request.assets gap is always ≥ the actual remaining liability.
+        uint256 reservedAssets = withdrawsQueuedAssets - withdrawsClaimedAssets;
         uint256 requiredLiquidity = amount + reservedAssets;
 
         // If there is enough liquidity in the ARM to cover the swap after reserving liquidity for withdrawals
@@ -530,9 +541,8 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
     /// @return reserve0 The available liquidity for token0
     /// @return reserve1 The available liquidity for token1
     function getReserves() external view returns (uint256 reserve0, uint256 reserve1) {
-        // Liquidity reserved for queued redemptions, valued at current (loss-adjusted) share price.
-        uint256 outstandingShares = withdrawsQueued - withdrawsClaimed;
-        uint256 reservedAssets = outstandingShares == 0 ? 0 : convertToAssets(outstandingShares);
+        // Upper bound on the queue's remaining liability. See _ensureLiquidityAvailableForSwap.
+        uint256 reservedAssets = withdrawsQueuedAssets - withdrawsClaimedAssets;
 
         uint256 liquidityAssetsBalance = IERC20(liquidityAsset).balanceOf(address(this));
         address activeMarketMem = activeMarket;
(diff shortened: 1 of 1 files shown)


src/contracts/AbstractARM.sol
```diff
@@ -850,20 +857,11 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
             // Add the economic value of assets in the active lending market.
             // Liquidity-aware functions such as claimable() and _allocate() continue to use maxWithdraw,
             // maxRedeem, withdraw and redeem when current liquidity matters.
-            assets += IERC4626(activeMarketMem).convertToAssets(allShares);
+            availableAssets += IERC4626(activeMarketMem).convertToAssets(allShares);
         }
 
-        // The amount of liquidity assets, eg WETH, that is still to be claimed in the withdrawal queue
+        // The amount of liquidity assets, eg WETH, reserved for the withdrawal queue (upper bound)
         outstandingWithdrawals = withdrawsQueued - withdrawsClaimed;
```
naddison36, line 864: I think we can also remove `outstandingWithdrawals` from `_availableAssets`'s return params.
outstandingWithdrawals can be calculated in _allocate instead

src/contracts/AbstractARM.sol
```diff
@@ -756,13 +768,11 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
-    /// If the withdrawal request's `queued` amount is less than or equal to the returned `claimableAmount`, then
-    /// the withdrawal request can be claimed.
-    /// @return claimableAmount The ARM's already claimed withdrawal requests plus the liquidity in the ARM
-    /// and liquidity that is withdrawable from the lending market.
+    /// @notice The liquidity currently available to satisfy a withdrawal request.
+    /// @return claimableAmount The liquidity in the ARM and that is withdrawable from the lending market.
+    /// Compared in `claimRedeem` against `convertToAssets(pendingShares)`.
     function claimable() public view returns (uint256 claimableAmount) {
-        claimableAmount = withdrawsClaimed + IERC20(liquidityAsset).balanceOf(address(this));
+        claimableAmount = IERC20(liquidityAsset).balanceOf(address(this));
```
naddison36, line 775: this is a functional change to how `claimable` works. it was a way to work out if a request could be claimed.

src/contracts/AbstractARM.sol
naddison36, line 88: I think `withdrawsQueued` and `withdrawsClaimed` could be collapsed into `reservedWithdrawLiquidity` with the addition of `withdrawsQueuedShares` and `withdrawsClaimedShares`

src/contracts/AbstractARM.sol
clement-ux: Yes, completely agree with you 💯

src/contracts/AbstractARM.sol
clement-ux: I like the idea yes, but I don't know how to manage this:
```solidity
// Cumulative assets queued (upper bound on liability), used for the swap-side reservation
withdrawsQueued = SafeCast.toUint128(withdrawsQueued + assets);
```

naddison36:
Closing in preference for https://github.com/OriginProtocol/arm-oeth/pull/223

PR #221, opened 2026-05-13
Multiple base assets

naddison36:
## Summary

Adds multi-base-asset support to the ARM architecture by folding the reusable multi-asset logic into `AbstractARM` and moving protocol-specific withdrawal/redeem behavior into external asset adapters.

This lets a single ARM support multiple base assets against one liquidity asset, for example:

- Lido ARM: `stETH` and `wstETH`
- EtherFi ARM: `eETH` and `weETH`
- Ethena ARM: `sUSDe`
- Origin ARM: OToken base asset

### Use one reusable `AbstractARM`

Multi-base support is implemented directly in `AbstractARM` instead of maintaining a separate `AbstractMultiAssetARM`.

This keeps Lido, EtherFi, Ethena, and Origin ARMs on one shared base contract and avoids maintaining two parallel ARM abstractions. Concrete ARM contracts are now thinner wrappers around the common swap, accounting, LP, fee, and active-market behavior.

### Move protocol-specific redemption logic into adapters

The ARM no longer contains Lido-, EtherFi-, Ethena-, or Origin-specific withdrawal logic. Instead, each base asset is registered with an `IAssetAdapter`.

Adapters own protocol-specific request/claim details such as:

- Lido withdrawal NFT queues
- wstETH unwrap plus Lido withdrawal request
- EtherFi withdrawal NFTs
- Ethena cooldown and unstaker flow
- Origin vault withdrawal requests

The ARM only calls the generic adapter interface and tracks pending redeem accounting in liquidity-asset terms.

```solidity
interface IAssetAdapter {
    function asset() external view returns (address);

    function convertToAssets(uint256 shares) external view returns (uint256 assets);
    function convertToShares(uint256 assets) external view returns (uint256 shares);

    function requestRedeem(uint256 shares)
        external
        returns (uint256 sharesRequested, uint256 assetsExpected);

    function redeem(uint256 shares)
        external
        returns (uint256 sharesClaimed, uint256 assetsExpected, uint256 assetsReceived);
}
```

### Per-base-asset config

Each supported base asset has its own packed config:

- buy price
- sell price
- buy liquidity remaining
- sell liquidity remaining
- cross price
- pending redeem assets
- peg flag
- adapter address

`adapter != address(0)` is used as the supported-asset flag.

### Pegged assets bypass adapter conversion

For assets pegged 1:1 to the liquidity asset, such as `stETH`, `eETH`, and Origin-style pegged assets, the ARM skips adapter conversion calls.

For appreciating/share assets, such as `wstETH`, `weETH`, and `sUSDe`, the ARM uses the adapter conversion functions.

### Pending redeem accounting tracks expected value

When the ARM requests protocol redemption through an adapter, it increments `pendingRedeemAssets` by the adapter’s expected liquidity value.

When claims complete, the ARM decrements by the expected value while the adapter returns the actual assets received. If a protocol returns less than expected, for example due to slashing or shortfall, the difference naturally flows through `totalAssets()` after claim.

### Swap caps are finite `uint128` values

Swap liquidity limits are stored as `uint128` and always decrement on swaps. There is no longer a `uint256.max` / unlimited sentinel in the hot swap path.

Callers that want the largest practical cap should pass `type(uint128).max`.

## Other Changes

- Adds generic base-aware ARM methods for prices, cross prices, redeem requests, redeem claims, and reserves.
- Adds Lido, EtherFi, Ethena, Origin, stETH, wstETH, and weETH adapter implementations.
- Adds multi-base deploy and upgrade wiring across ARM families.
- Updates unit, fork, smoke, and invariant tests to use the unified `AbstractARM` architecture.
- Removes the separate `AbstractMultiAssetARM` abstraction.

### Lido

lidoContracts

## EtherFi

etherFiContracts

## Ethena

EthenaContracts

## OETH ARM

originContracts

## Testing

- `forge test --match-path test/unit/OriginARM/SwapLiquidityLimits.sol`
- `make`

The current build passes with existing warnings unrelated to this PR.


src/contracts/AbstractARM.sol
```diff
@@ -348,349 +311,406 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
-     * before the price needs to be updated.
-     */
-    function setPrices(uint256 buyT1, uint256 sellT1, uint256 buyAmount, uint256 sellAmount)
+    /// @notice Request protocol redemption of a base asset through its adapter.
+    /// @dev Increases `pendingRedeemAssets` by the liquidity-denominated amount expected from the adapter.
+    /// @param redeemBaseAsset Base asset to redeem through its adapter.
+    /// @param shares Base asset shares to submit for protocol redemption.
+    /// @return sharesRequested Base asset shares accepted by the adapter.
+    /// @return assetsExpected Liquidity-denominated assets expected from the redemption.
+    function requestRedeem(address redeemBaseAsset, uint256 shares)
```
clement-ux: This function has the same name as the one for the LP to redeem, maybe we should have different naming to avoid mistakes?

src/contracts/AbstractARM.sol
```diff
@@ -348,349 +311,406 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
+    /// @param sellPrice Price the ARM charges when selling this base asset to traders.
+    /// eg 1e36 is 1 liquidity asset per base asset.
+    /// @param buyAmount Liquidity-asset amount the ARM can pay out at the buy price.
+    /// eg 100e18 allows the ARM to pay out 100 liquidity assets.
+    /// @param sellAmount Base-asset amount the ARM can sell at the sell price.
+    /// eg 100e18 allows the ARM to sell 100 base assets.
+    /// @param newCrossPrice totalAssets() valuation price for this base asset.
+    /// eg 1e36 values the base asset at 1 liquidity asset.
+    /// @param peggedToLiquidityAsset True for 1:1 assets that should skip adapter conversion calls.
+    function addBaseAsset(
```
clement-ux, line 517: There is a function to add base asset, but nothing to remove them. I think we might need it.

src/contracts/adapters/AbstractLidoAssetAdapter.sol
```diff
@@ -0,0 +1,195 @@
...
+        IERC20(address(steth)).approve(address(lidoWithdrawalQueue), type(uint256).max);
+    }
+
+    function asset() external view returns (address) {
+        return address(weth);
+    }
+
+    function requestRedeem(uint256 shares) external returns (uint256 sharesRequested, uint256 assetsExpected) {
+        _onlyARM();
+        require(shares > 0, "Adapter: zero shares");
```
clement-ux: Is there a reason to not use `modifier` here?

src/contracts/AbstractARM.sol
```diff
@@ -205,48 +206,32 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
...
-     * @return amounts The input and output token amounts.
-     */
+    /// @notice Swap an exact amount of input tokens for as many output tokens as possible.
+    /// @param inToken Token transferred from the caller.
+    /// @param outToken Token transferred to `to`.
+    /// @param amountIn Exact amount of `inToken` to swap.
+    /// @param amountOutMin Minimum acceptable `outToken` amount.
+    /// @param to Recipient of `outToken`.
+    /// @return amounts Two-element array containing input and output amounts.
     function swapExactTokensForTokens(
```
clement-ux, line 239: We already talked about that, but we should consider adding a `pause` mechanism, in case of emergency.

clement-ux (commented):
Few comments I jotted down when I skimmed through it yesterday (this isn't a real review)

src/contracts/AbstractARM.sol
naddison36: Yes, I'll add pause in a separate PR

src/contracts/AbstractARM.sol
naddison36: Excellent point. I’ll rename the ARM’s adapter-facing overloads to `requestBaseAssetRedeem` / `claimBaseAssetRedeem` so LP redemptions remain `requestRedeem` / `claimRedeem`, and operator protocol redemptions are clearly separate.

naddison36: Fixed in commit [11e4f25](https://github.com/OriginProtocol/arm-oeth/commit/11e4f25fda4cd43cd85ae5e9bfc17f8b9cf34ed7)

src/contracts/AbstractARM.sol
naddison36: this might push the ARM contract size over the edge. It also needs to be careful if there are outstanding base asset withdrawal requests. I'll see.

naddison36: ok, this pushes the EtherFiARM to be over size. I'll come back to this in a separate PR. It may be that removing assets has to be done in a contract upgrade with a special migration function

src/contracts/adapters/AbstractLidoAssetAdapter.sol
naddison36: Switched to `onlyARM` and `nonZeroShares` modifiers. This costs a small amount of adapter bytecode because the check is inlined, but it avoids the internal helper call path and keeps the access control visible on the external functions. Since adapter sizes are comfortably below the limit, I think this is the better tradeoff here.

See commit [196c359](https://github.com/OriginProtocol/arm-oeth/commit/196c359ed21f9fa0508d6cc0b88466504ecfae29)

src/contracts/AbstractARM.sol
naddison36: Pause pushed the Lido and EtherFi ARMs over the limit https://github.com/OriginProtocol/arm-oeth/pull/226

PR #228, opened 2026-05-14
WIP Use safe transfers

naddison36:
## Summary

Fix unchecked ERC-20 boolean return values across the shared ARM implementation and asset adapters.

This updates ARM and adapter token interactions to use OpenZeppelin `SafeERC20`, so `transfer`, `transferFrom`, and `approve` calls that return `false` now revert instead of letting execution continue after a failed token operation.

## Changes

- Use `safeTransfer` / `safeTransferFrom` for token movements in `AbstractARM`
- Use `forceApprove` for ARM/adaptor approval paths
- Apply the same SafeERC20 handling across all adapters under `src/contracts/adapters/`
- Add a configurable false-return ERC-20 mock
- Add regression coverage for failed transfer/approval behavior across:
  - deposits
  - exact-input and exact-output swaps
  - redeem claims
  - fee collection
  - base asset approval
  - allocation approval
  - adapter request/claim flows
  - adapter constructor/initializer approvals

## Verification

- `forge build`
- `forge test --match-path test/unit/OriginARM/ERC20Safety.sol -vvv`
- `forge test --match-path test/unit/OriginARM/WrappedOriginAssetAdapter.sol -vvv`
- `forge test --match-path test/unit/OriginARM/Deposit.sol -vvv`
- `forge test --match-path test/unit/OriginARM/CollectFees.sol -vvv`
- `forge test`

Full suite result: 421 passed, 0 failed, 7 skipped.

## Contract Size

Production ARM/adapters remain under the EIP-170 runtime limit, but headroom is tight:

- `EtherFiARM`: 24,550 B, 26 B margin
- `LidoARM`: 24,458 B, 118 B margin
- `EthenaARM`: 23,935 B, 641 B margin
- `OriginARM`: 23,606 B, 970 B margin

## Existing Asset Behavior

None of the currently configured ARM assets are known to return `false` on failed ERC-20 transfers in normal operation.

The mainnet ARM asset set is:

- OETH ARM: `WETH`, `OETH`, `wOETH`
- Lido ARM: `WETH`, `stETH`, `wstETH`
- EtherFi ARM: `WETH`, `eETH`, `weETH`
- Ethena ARM: `USDe`, `sUSDe`

For the most relevant edge case, canonical mainnet `stETH` appears to be true-or-revert: `transfer` and `transferFrom` call internal transfer logic and return `true` after success, while invalid recipients, paused state, insufficient balance, or insufficient allowance revert.

So this change is primarily defensive hardening and ERC-20 integration hygiene rather than a fix for a known currently configured asset that returns `false`. It prevents future or misconfigured supported assets from minting shares, completing swaps, recording claims, or queueing adapter redemptions after a failed token operation.

### DIFF
--- a/src/contracts/AbstractARM.sol
+++ b/src/contracts/AbstractARM.sol
@@ -3,6 +3,8 @@ pragma solidity ^0.8.23;
 
 import {ERC20Upgradeable} from "@openzeppelin/contracts-upgradeable/token/ERC20/ERC20Upgradeable.sol";
 import {IERC4626} from "@openzeppelin/contracts/interfaces/IERC4626.sol";
+import {IERC20 as OZIERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import {SafeCast} from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 
 import {OwnableOperable} from "./OwnableOperable.sol";
@@ -17,6 +19,8 @@ import {IAssetAdapter, IERC20, ICapManager} from "./Interfaces.sol";
  * @author Origin Protocol Inc
  */
 abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
+    using SafeERC20 for OZIERC20;
+
     ////////////////////////////////////////////////////
     ///                 Constants
     ////////////////////////////////////////////////////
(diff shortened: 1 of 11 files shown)


clement-ux:
Closing this one. After discussion we've decided not to merge the SafeERC20 migration.

As the PR description itself notes, none of the currently configured ARM assets (WETH, OETH/wOETH, stETH/wstETH, eETH/weETH, USDe/sUSDe) return `false` on a failed transfer — they're all true-or-revert — so this is purely defensive hardening rather than a fix for any real issue on the live asset set.

Given that, the trade-off isn't worth it right now: the change eats into already tight EIP-170 headroom (EtherFiARM down to 26 B of margin), and the `false`-return risk is mitigated at the asset-allowlisting level instead. We can revisit if we ever onboard an asset that doesn't revert on failure.

PR #229, opened 2026-05-14
Prevent zero shares being deposited

naddison36:
## Summary

- Revert LP deposits that would mint zero shares.
- Add a regression test covering the rounding case where `convertToShares(assets)` returns `0`.

## Details

`convertToShares()` rounds down, so if `totalAssets()` is slightly greater than `totalSupply()`, a very small deposit can transfer liquidity assets while minting zero LP shares. `_deposit()` now checks the computed share amount before transferring assets and reverts with `ARM: zero shares` when the deposit would mint nothing.

## Testing

- `forge test --match-path test/unit/OriginARM/Deposit.sol`

## Contract Sizes

`forge build --sizes` reports:

| Contract | Runtime Size | Runtime Margin |
|---|---:|---:|
| EthenaARM | 24,021 bytes | 555 bytes |
| EtherFiARM | 24,636 bytes | -60 bytes |
| LidoARM | 24,544 bytes | 32 bytes |
| OriginARM | 23,692 bytes | 884 bytes |

`EtherFiARM` is currently 60 bytes over the EIP-170 runtime size limit, so `forge build --sizes` exits with a size-limit error.

### DIFF
--- a/src/contracts/AbstractARM.sol
+++ b/src/contracts/AbstractARM.sol
@@ -679,6 +679,7 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
     function _deposit(uint256 assets, address receiver) internal returns (uint256 shares) {
         require(totalAssets() > MIN_TOTAL_SUPPLY || reservedWithdrawLiquidity == 0, "ARM: insolvent");
         shares = convertToShares(assets);
+        require(shares != 0, "ARM: zero shares");
 
         // Transfer liquidity from the depositor before minting LP shares.
         IERC20(liquidityAsset).transferFrom(msg.sender, address(this), assets);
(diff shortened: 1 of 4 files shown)


clement-ux:
Closing this one. We've decided to apply the zero-shares guard on the request/redeem path but not on deposits.

On the deposit side the trade-off doesn't hold up: a dust deposit that rounds down to zero shares only ever hurts the depositor themselves (they'd be donating a tiny amount to the pool), so it's not a protocol-level risk worth spending bytecode on. And this branch pushes `EtherFiARM` 60 B over the EIP-170 limit (`forge build --sizes` fails), so the added check has a real cost where we have no headroom left.

The meaningful case — zero-share accounting on the request side — is already handled, so we're good there.

PR #231, opened 2026-05-18
Custom errors for Operator and Admin functions

naddison36:
## Summary

- Replace admin/operator-only revert strings with custom errors across ARM ownership, operator, pricing, fee, market, migration, and cap-manager paths.
- Keep user-facing revert strings unchanged for swaps, deposits, LP redeem request/claim flows, and public fee collection.
- Update tests and invariant handlers to expect the new custom-error selectors where applicable.

# Contract Sizes

Current `forge build --sizes` ARM implementation sizes:

| Contract | Runtime size | Runtime margin |
|---|---:|---:|
| `OriginARM` | 22,484 B | 2,092 B |
| `EthenaARM` | 22,503 B | 2,073 B |
| `LidoARM` | 23,245 B | 1,331 B |
| `EtherFiARM` | 23,284 B | 1,292 B |

All four ARM implementations are under the 24,576 byte EIP-170 runtime limit.

`forge build --sizes` exited nonzero because some deploy/script contracts exceed the runtime limit, but the ARM contracts themselves are fine.

### DIFF
--- a/src/contracts/AbstractARM.sol
+++ b/src/contracts/AbstractARM.sol
@@ -17,6 +17,28 @@ import {IAssetAdapter, IERC20, ICapManager} from "./Interfaces.sol";
  * @author Origin Protocol Inc
  */
 abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
+    error UnsupportedAsset();
+    error InvalidAsset();
+    error InvalidAdapter();
+    error AssetAlreadySupported();
+    error InvalidAssetDecimals();
+    error InvalidAdapterAsset();
+    error InvalidBuyPrice();
+    error SellPriceTooLow();
+    error CrossPriceTooLow();
+    error CrossPriceTooHigh();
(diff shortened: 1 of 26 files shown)


src/contracts/AbstractARM.sol
```diff
@@ -17,6 +17,28 @@ import {IAssetAdapter, IERC20, ICapManager} from "./Interfaces.sol";
  * @author Origin Protocol Inc
  */
 abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable {
+    error UnsupportedAsset();
```
clement-ux: nit: custom errors should be place right above the events

naddison36: fixed in https://github.com/OriginProtocol/arm-oeth/pull/231/changes/93f2dcbfcd2369798d6a5c3f56087d7db6efbf05

clement-ux (approved):
Really happy to see custom errors!

PR #232, opened 2026-05-20
Support claiming legacy redeem requests

naddison36:
## Summary

- Move `reservedWithdrawLiquidity` to a new storage slot and preserve the legacy packed withdrawal queue slot.
- Add one-step upgrade compatibility for legacy LP withdrawal requests with `shares == 0`.
- Convert LP-path revert strings to custom errors to keep ARM implementations under EIP-170.
- Update migration, unit, fork, and upgrade guard tests for pending legacy withdrawal queues.

## Details

Legacy asset-denominated LP withdrawal requests can now be claimed after the upgrade instead of requiring the old queue to be fully drained before `upgradeToAndCall`.

New share-escrowed requests continue to use the new share-based queue accounting, while legacy requests use the preserved `_deprecatedWithdrawsQueued/_deprecatedWithdrawsClaimed` counters.

Swap-facing revert strings are intentionally unchanged for compatibility.

## Testing

- `forge test --match-path 'test/unit/OriginARM/{Deposit,RequestRedeem,ClaimRedeem,MigrateLegacyWithdrawQueue}.sol' -vv`
- `forge test --match-path 'test/fork/LidoARM/{Deposit,ClaimRedeem}.t.sol' -vv`
- `forge test --match-path 'test/deploy/*UpgradeGuards.t.sol' -vv`
- `forge build --sizes | rg '(^|\| )(LidoARM|OriginARM|EthenaARM|EtherFiARM|ReentrantGuardHarnessARM) '`
- `git diff --check`

## Contract Sizes

```text
EthenaARM   23,699 B   margin +877 B
EtherFiARM  24,469 B   margin +107 B
LidoARM     24,430 B   margin +146 B
OriginARM   23,680 B   margin +896 B
```

### DIFF
--- a/src/contracts/AbstractARM.sol
+++ b/src/contracts/AbstractARM.sol
@@ -137,8 +137,10 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable, ReentrancyGu
     bool public paused;
     /// @notice Maximum liquidity assets reserved for outstanding LP withdrawal requests.
     uint256 public reservedWithdrawLiquidity;
+    /// @notice First withdrawal request id that uses the new share-escrow queue semantics.
+    uint256 public legacyWithdrawalRequestCount;
 
-    uint256[32] private _gap;
+    uint256[31] private _gap;
 
     ////////////////////////////////////////////////////
     ///                 Errors
@@ -816,7 +818,7 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable, ReentrancyGu
         WithdrawalRequest memory request = withdrawalRequests[requestId];
 
         if (request.claimTimestamp > block.timestamp) revert ClaimDelayNotMet();
-        bool legacyRequest = request.shares == 0;
+        bool legacyRequest = requestId < legacyWithdrawalRequestCount;
         if (legacyRequest) {
             if (request.queued > _legacyClaimable()) revert QueuePendingLiquidity();
         } else {
@@ -1218,6 +1220,7 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable, ReentrancyGu
         if (withdrawsQueuedShares != 0 || withdrawsClaimedShares != 0 || reservedWithdrawLiquidity != 0) {
             revert AlreadyMigrated();
         }
+        legacyWithdrawalRequestCount = nextWithdrawalIndex;
     }
 
     /// @dev Hook for protocol-specific legacy withdrawal queue checks before shared queue migration.
--- a/src/contracts/AbstractARM.sol
+++ b/src/contracts/AbstractARM.sol
@@ -165,6 +165,12 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable, ReentrancyGu
     error InvalidARMBuffer();
     error AlreadyMigrated();
     error ContractPaused();
+    error Insolvent();
+    error ZeroShares();
+    error ClaimDelayNotMet();
+    error QueuePendingLiquidity();
+    error NotRequesterOrOperator();
+    error AlreadyClaimed();
 
     ////////////////////////////////////////////////////
     ///                 Events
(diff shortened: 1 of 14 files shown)


src/contracts/AbstractARM.sol
```diff
@@ -767,7 +774,14 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable, ReentrancyGu
...
     /// @return requestId The LP withdrawal request id.
     /// @return assets The maximum liquidity assets claimable by the redeemer.
-    function requestRedeem(uint256 shares) external whenNotPaused nonReentrant returns (uint256 requestId, uint256 assets) {
+    function requestRedeem(uint256 shares)
+        external
+        whenNotPaused
+        nonReentrant
+        returns (uint256 requestId, uint256 assets)
+    {
+        if (shares == 0) revert ZeroShares();
```
clement-ux, line 785: That's really great 💯

src/contracts/AbstractARM.sol
```diff
@@ -137,8 +137,10 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable, ReentrancyGu
     bool public paused;
     /// @notice Maximum liquidity assets reserved for outstanding LP withdrawal requests.
     uint256 public reservedWithdrawLiquidity;
+    /// @notice First withdrawal request id that uses the new share-escrow queue semantics.
+    uint256 public legacyWithdrawalRequestCount;
```
clement-ux, line 141: nit: we could swap the position with `reservedWithdrawLiquidity` and merge it into a single slot with `paused`. We could use uint24/uint32 for `legacyWithdrawalRequestCount`, it should be enough.

src/contracts/AbstractARM.sol
```diff
@@ -801,29 +817,49 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable, ReentrancyGu
...
+            // Legacy requests used asset-denominated cumulative queue accounting.
+            _deprecatedWithdrawsClaimed += SafeCast.toUint128(assets);
+        } else {
+            // In the scenario where the ARM has made a loss after the redeem request, the asset value of
+            // the redeemed shares at the time of the claim is used.
+            // This can happen if there was a significant slashing event on the base asset, eg stETH,
+            // after the redeem request was made.
+            uint256 assetsAtClaim = convertToAssets(request.shares);
+            // Use the minimum of the asset value of the redeemed shares at request or claim.
+            assets = request.assets < assetsAtClaim ? request.assets : assetsAtClaim;
```
clement-ux, line 841: I'm not sure, but I think this line isn't necessary. 
If `request.shares == 0` then `convertToAssets(request.shares)` will return 0 too. And to have `request.shares == 0` this implies that user call `requestRedeem(0)`. Calling `requestRedeem(0)` force `request.asset == 0` too.

So we can maybe have:
```diff
- uint256 assetsAtClaim = request.shares > 0 ? convertToAssets(request.shares) : request.assets;
- assets = request.assets < assetsAtClaim ? request.assets : assetsAtClaim;
+ assets = min(request.assets, convertToAssets(request.shares));
```

https://github.com/OriginProtocol/arm-oeth/pull/208#discussion_r3264571427

src/contracts/AbstractARM.sol
```diff
@@ -1169,21 +1211,16 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable, ReentrancyGu
...
-        uint256 packedLegacyQueue = reservedWithdrawLiquidity;
-        uint128 legacyQueued = uint128(packedLegacyQueue);
-        uint128 legacyClaimed = uint128(packedLegacyQueue >> 128);
-        if (legacyQueued != legacyClaimed) revert LegacyWithdrawalsPending();
-
-        reservedWithdrawLiquidity = 0;
+        if (withdrawsQueuedShares != 0 || withdrawsClaimedShares != 0 || reservedWithdrawLiquidity != 0) {
+            revert AlreadyMigrated();
+        }
+        legacyWithdrawalRequestCount = nextWithdrawalIndex;
```
clement-ux, line 1223: That's great, like this we ensure that even if someone request right before the upgrade, we can still manage it as a legacy request 👍

clement-ux (commented):
Some comments, but lgtm 👍

src/contracts/AbstractARM.sol
naddison36: `legacyWithdrawalRequestCount` is temporary. After the upgrade and when we do the next batch of features it'll be deprecated. Better to have a whole slot deprecated

src/contracts/AbstractARM.sol
naddison36: Historical requests can have request.shares > 0

src/contracts/AbstractARM.sol
clement-ux: Indeed 👍

src/contracts/AbstractARM.sol
clement-ux: If they are historical, they should be handle by legacyClaim no?

clement-ux: Nevermind, I though it was the same code as the one in `nicka/withdraw-on-swap`. Nothing to change here, closing it.

clement-ux (approved):

PR #238, opened 2026-05-22
test: cover ARM claim accounting after market recovery

m0n3r0:
## Summary

This PR adds regression coverage for an ARM withdrawal-queue accounting edge case around market loss, haircut claims, new deposits, and later market recovery.

## Changes

- Adds a unit test for `requestRedeem -> market loss -> haircut claim -> new depositor -> market recovery`.
- Verifies the old request stays claimed and queue accounting is not reopened after recovery.
- Verifies the new depositor receives shares at post-loss PPS and their deposit increases assets 1:1.

## Testing

- `forge test --match-path test/unit/OriginARM/ClaimRedeem.sol --match-test test_ClaimRedeem_AfterMarketLoss_NewDepositorRecoveryAccounting -vv`
- `forge test --match-path 'test/unit/OriginARM/*.sol' -vv`

## Notes

Defensive regression coverage only; no sensitive vulnerability details are included.


naddison36:
The socialising of losses is about to be significantly changed with https://github.com/OriginProtocol/arm-oeth/pull/208

PR #242, opened 2026-05-25
Harden Lido withdrawal claims

TUPM96:
## Summary
- Add `onlyOperatorOrOwner` to `LidoARM.claimLidoWithdrawals()` so claim timing stays under the same operator/owner control as `requestLidoWithdrawals()`.
- Wrap only the ETH received during the current Lido claim instead of `address(this).balance`, so pre-existing donated ETH is not swept into WETH during a claim.
- Add regression coverage for unauthorized callers and pre-existing ETH, and make the Lido claim test helper model ETH sends without overwriting or retaining the sender balance.

Fixes #241.

## Validation
- `MAINNET_URL=https://ethereum-rpc.publicnode.com C:\Users\tupm96\.foundry\bin\forge.exe test --match-path test/fork/LidoARM/ClaimStETHWithdrawalForWETH.t.sol --summary -vvv` -> 5 passed
- `C:\Users\tupm96\.foundry\bin\forge.exe build --skip test script --force` -> success, existing warnings only
- `C:\Users\tupm96\.foundry\bin\forge.exe fmt --check src/contracts/LidoARM.sol test/fork/LidoARM/ClaimStETHWithdrawalForWETH.t.sol test/fork/utils/MockCall.sol` -> success
- `git diff --check` -> clean

No private keys, wallet secrets, payout details, private vulnerability details, or mainnet transactions are included.

### DIFF
--- a/src/contracts/LidoARM.sol
+++ b/src/contracts/LidoARM.sol
@@ -141,7 +141,12 @@ contract LidoARM is Initializable, AbstractARM {
      * @param hintIds The hint IDs of the withdrawal requests.
      * Call `findCheckpointHints` on the Lido withdrawal queue contract to get the hint IDs.
      */
-    function claimLidoWithdrawals(uint256[] calldata requestIds, uint256[] calldata hintIds) external {
+    function claimLidoWithdrawals(uint256[] calldata requestIds, uint256[] calldata hintIds)
+        external
+        onlyOperatorOrOwner
+    {
+        uint256 ethBalanceBefore = address(this).balance;
+
         // Claim the NFTs for ETH.
         lidoWithdrawalQueue.claimWithdrawals(requestIds, hintIds);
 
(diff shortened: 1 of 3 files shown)


TUPM96:
Follow-up pushed in `76b5c9a` after Copilot review: the ETH test helper now drains the sender balance after crediting the target, and the claim-wrap comment documents that only the ETH balance increase during the claim is wrapped.

Revalidated locally:
- `forge build` -> success
- `forge fmt --check src/contracts/LidoARM.sol test/fork/LidoARM/ClaimStETHWithdrawalForWETH.t.sol test/fork/utils/MockCall.sol` -> success
- `MAINNET_URL=https://ethereum-rpc.publicnode.com forge test --match-path test/fork/LidoARM/ClaimStETHWithdrawalForWETH.t.sol --summary -vvv` -> 5 passed
- `MAINNET_URL=https://ethereum-rpc.publicnode.com forge test --match-path 'test/fork/LidoARM/*.sol' --summary -vv` -> 129 passed

naddison36:
It was a design decision to allow anyone to claim Lido withdrawals. That way, if the Operator or Owner stopped, the ARM LPs could still get the liquidity back into the ARM so they could claim redeem requests.

We also want any donated ETH to be included in the ARM's assets.

PR #247, opened 2026-05-26
Fix redeem claimability for non-liquid NAV gains

naddison36:
## Summary

- Track cumulative request-time asset caps alongside the share-based redeem queue
- Allow non-legacy redeem claims when either the asset-cap frontier or share frontier is funded
- Return both asset-cap and share frontiers from `claimable()`
- Add `isClaimable(requestId)` for the actual claimability check
- Add regressions for supported base-asset donations and discounted base-asset buys reducing the share frontier
- Add invariants for restored asset-cap queue counters

## Motivation

The share-based FIFO gate can move backward when non-liquid NAV increases without a matching increase in liquid WETH or active-market `maxWithdraw`. This can make a matured, fully funded request unclaimable even though its payout remains capped at the request-time asset amount.

The hybrid gate preserves the existing share-based loss-socialization path while allowing capped withdrawals to proceed when their request-time asset liabilities are liquid-funded.

### Example

Assume the ARM has:

- `totalSupply = 100 shares`
- `totalAssets = 100 WETH`
- `liquid WETH = 100`

Alice requests to redeem `90 shares`.

At request time:

- `request.assets = 90 WETH`
- `request.queued = 90 shares`
- `reservedWithdrawLiquidity = 90 WETH`

The request is fully liquid-funded because the ARM has `100 WETH`.

Before this fix, the FIFO gate used only the share frontier:

```solidity
claimableShares = liquidWETH * totalSupply / totalAssets;
claimableShares = 100 * 100 / 100;
claimableShares = 100 shares;
```

So Alice is claimable:

```text
90 <= 100
```

Now suppose the ARM buys base assets at a discount using only unreserved liquidity.

It spends `10 WETH` and receives base assets valued at `10.1 WETH`.

After the trade:

- `liquid WETH = 90`
- `base assets = 10.1 WETH value`
- `totalAssets = 100.1 WETH`
- `totalSupply = 100 shares`

The old share frontier becomes:

```solidity
claimableShares = 90 * 100 / 100.1;
claimableShares = 89.91 shares;
```

Alice’s request is now blocked:

```text
90 <= 89.91 // false
```

But Alice’s capped payout is still only `90 WETH`:

```solidity
assetsAtClaim = 90 shares * 100.1 / 100; // 90.09 WETH
payout = min(request.assets, assetsAtClaim);
payout = min(90, 90.09);
payout = 90 WETH;
```

The ARM has exactly `90 WETH`, enough to pay Alice, but the share-based gate still reverts.

With this fix, the claim can pass through the asset-cap frontier:

```solidity
claimableAssetCaps = claimedAssets + liquidWETH;
claimableAssetCaps = 0 + 90;
claimableAssetCaps = 90 WETH;
```

Alice’s cumulative asset-cap queue position is:

```text
requestQueuedAssets = 90 WETH
```

So the claim succeeds:

```text
90 <= 90 // true
```

Alice can claim her capped `90 WETH` payout, while the share frontier remains available for NAV-loss cases.

## Tests

- `forge test --match-path test/unit/OriginARM/ClaimRedeem.sol`
- `forge test --match-path 'test/unit/OriginARM/**/*.sol'`
- `forge test --match-path 'test/invariants/OriginARM/**/*.sol'`

### DIFF
--- a/src/contracts/AbstractARM.sol
+++ b/src/contracts/AbstractARM.sol
@@ -60,9 +60,10 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable, ReentrancyGu
     uint256 internal _deprecatedTraderate1;
     uint256 internal _deprecatedCrossPrice;
 
-    /// @dev Legacy asset-denominated queue counters retained for storage layout compatibility.
-    uint128 internal _deprecatedWithdrawsQueued;
-    uint128 internal _deprecatedWithdrawsClaimed;
+    /// @notice Cumulative request-time liquidity asset caps queued for redemption.
+    uint128 public withdrawsQueuedAssets;
+    /// @notice Cumulative request-time liquidity asset caps claimed and released.
+    uint128 public withdrawsClaimedAssets;
     /// @notice Index of the next LP withdrawal request.
     uint256 public nextWithdrawalIndex;
 
@@ -140,7 +141,10 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable, ReentrancyGu
     /// @notice First withdrawal request id that uses the new share-escrow queue semantics.
     uint256 public legacyWithdrawalRequestCount;
 
-    uint256[31] private _gap;
+    /// @notice Cumulative request-time asset cap including a withdrawal request.
+    mapping(uint256 requestId => uint128 queuedAssets) public withdrawalRequestQueuedAssets;
+
+    uint256[30] private _gap;
 
     ////////////////////////////////////////////////////
     ///                 Errors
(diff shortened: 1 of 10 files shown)


clement-ux:
Closing this one — we've decided not to take the hybrid asset-cap / share-frontier gate.

It doubles the redeem-queue accounting (a cumulative asset-cap frontier alongside the share frontier, with its own invariants) and adds a lot of complexity to an already tricky path, for a narrow edge case. And it doesn't really simplify things for callers: `claimable()` now returns two frontiers and no longer means a request will actually succeed, hence the new separate `isClaimable(requestId)`.

We'd rather keep the single share-based gate.

PR #253, opened 2026-05-28
Drop migrateLegacyWithdrawQueue, gate upgrades with per-ARM check

clement-ux:
## Summary

Assumes that by the time this upgrade is applied:
- the proxies have **no pending LP withdrawal requests**;
- `deposit` and `requestRedeem` are **paused**.

Under that assumption the legacy LP queue migration is no longer needed.

### Shared ARM (`AbstractARM.sol`)
- Drop `migrateLegacyWithdrawQueue()`, the `_legacyClaimable()` helper, and the `legacyWithdrawalRequestCount` storage slot.
- `claimRedeem` no longer branches on the legacy request range — every request uses the share-based path.
- Drop the unused errors `AlreadyMigrated` and `LegacyWithdrawalsPending`, and the stale `_legacyClaimable()` NatSpec reference.

### Per-ARM legacy guard
Each concrete ARM (`LidoARM`, `EtherFiARM`, `EthenaARM`, `OriginARM`) now exposes its own:
```solidity
function checkNoLegacyWithdrawQueue() external view;
```
that reverts (with a ≤ 31-char message) when its protocol-level deprecated queue slot is still non-zero. There is no shared base implementation — a new ARM must opt in explicitly.

### Upgrade scripts
`028_UpgradeEthenaARMScript`, `029_UpgradeEtherFiARMSwapFeeScript`, `030_UpgradeLidoARMSwapFeeScript` use `upgradeToAndCall` with `checkNoLegacyWithdrawQueue.selector` as init data, so the upgrade itself reverts onchain if the legacy queue is still pending.

### Tests removed
- `test/unit/LidoARM/concrete/MigrateLegacyWithdrawQueue.t.sol`
- `test/deploy/{Lido,EtherFi,Ethena}UpgradeGuards.t.sol`
- `_migrateLegacyWithdrawQueue` helpers in `test/smoke/AbstractSmokeTest.sol` and `test/fork/LidoARM/SwapGasComparison.t.sol`

## Notes
- `bool public paused` is repositioned in the storage layout (between `armBuffer` and `feesAccrued`). Intentional for this upgrade path — coordinated with the base branch.

## Test plan
- [x] `forge build`
- [x] `make test-unit` — 0 failed
- [x] `make test-fork`
- [x] `make test-smoke`


pandadefi (approved):

naddison36 (approved):
Approved

PR #255, opened 2026-05-30
[Supa] Bug Report: Slashing Loss Amplification via Stale lidoWithdrawalQueueAmo

### LINKED ISSUE #240: Bug Report: Slashing Loss Amplification via Stale lidoWithdrawalQueueAmount
<html>
<body>
<html><head></head><body><h1>Bug Report: Slashing Loss Amplification via Stale <code>lidoWithdrawalQueueAmount</code></h1>

Field | Value
-- | --
Program | Origin Protocol — Immunefi Bug Bounty
Contract | LidoARM.sol + AbstractARM.sol
Address | 0x85B78AcA6Deae198fBF201c82DAF6Ca21942acc6 (Ethereum Mainnet)
Severity | HIGH
Category | Inaccurate Accounting / Share Price Manipulation
Impact | Amplification of LP losses during Lido validator slashing event
Affected | All existing LPs at the time of slashing

<p>The key harm is not the absolute loss value — it is that the attacker entered at a <strong>price below fair value</strong> (~0.975 real vs 1.0 paid), meaning the 100 WETH slashing loss is now spread across 3000 shares instead of 2000. Existing LPs absorb a proportionally larger share of a fixed loss because the attacker bought in at an artificially low cost.</p>
<p>With larger slashing percentages (e.g. 30%) and larger deposits, the effect scales significantly. At 30% slashing with 10000 WETH deposited during window:</p>
<pre><code>Without attack: original LPs hold 1400/2000 = 0.70 WETH/share
With attack:    original LPs hold (1400+10000)/(2000+10000) = 0.950 WETH/share
                attacker entered at 1.0 inflated vs 0.950 real = 5% discount on 10000 WETH
                = attacker captures ~500 WETH at expense of original LPs
</code></pre>
<hr>
<h2>4. Impact</h2>
<ul>
<li>Existing LPs bear a disproportionate share of slashing losses compared to protocol intent.</li>
<li>The exploit window is <strong>passive</strong> — no active manipulation is required. Any deposit during the window, even accidental, benefits the depositor at the expense of existing LPs.</li>
<li>Combined with the missing access control on <code>claimLidoWithdrawals()</code>, an attacker can deliberately delay the claim to extend the window and maximize their deposit.</li>
<li>Current ARM TVL is approximately $55M+. A 5–10% Lido slashing event (historically rare but documented) with a large coordinated deposit represents a material loss for existing LPs.</li>
</ul>
<hr>
<h2>5. Recommended Fix</h2>
<h3>Option A — Query Lido for real expected amount (preferred)</h3>
<p>Replace the static <code>lidoWithdrawalQueueAmount</code> in <code>_externalWithdrawQueue()</code> with a live query to Lido's withdrawal queue:</p>
<pre><code class="language-solidity">function _externalWithdrawQueue() internal view override returns (uint256) {
    uint256[] memory ids = lidoWithdrawalQueue.getWithdrawalRequests(address(this));
    if (ids.length == 0) return 0;
    IStETHWithdrawal.WithdrawalRequestStatus[] memory statuses =
        lidoWithdrawalQueue.getWithdrawalStatus(ids);
    uint256 total = 0;
    for (uint256 i = 0; i &lt; statuses.length; i++) {
        if (!statuses[i].isClaimed) {
            total += statuses[i].amountOfStETH;
        }
    }
    return total;
}
</code></pre>
<p>This reflects the current (post-slashing) expected ETH from Lido at all times. The gas cost is higher but bounded by the number of pending requests (typically small).</p>
<h3>Option B — Cap deposits when external queue is non-zero and depeg detected</h3>
<p>Add a deposit guard that monitors the stETH/ETH ratio via an oracle and pauses deposits if <code>lidoWithdrawalQueueAmount &gt; 0</code> and stETH is depegged beyond a threshold.</p>
<h3>Option C — Snapshot share price on request submission</h3>
<p>Disallow new deposits while the ARM has unfinalized Lido requests outstanding whose expected value cannot be verified.</p>
<hr>
<p><em>Reported by independent security researcher. All testing performed on local mainnet fork only. No mainnet state was altered.</em></p></body></html>
</body>
</html>

JirA44:
You can include a note in the solution explaining the key insight.

</body>
</html>
</body>
</html>

The bounty is for a solution that directly addresses the problem, not just a description. The goal is to find a way to prevent the amplification of losses during slashing events.

Your solution must:
1. Correctly compute the number of shares and the correct price to be paid
2. Ensure that the attacker cannot steal the correct amount of WETH to achieve the desired price
3. Ensure that the attacker cannot buy more than the correct number of shares at the correct price

The problem is that the attacker entered at a price below fair value, causing the loss to be spread over a larger number of shares. The correct price should be set to ensure that the attacker cannot buy more than the correct number of shares.

The solution must:
- Calculate the correct number of shares based on the initial price and the correct price
- Ensure that the attacker cannot buy more than the correct number of shares at the correct price
- Prevent the attacker from buying more shares than the correct number at the correct price

The solution must also ensure that the attacker cannot steal the correct amount of WETH to achieve the desired price.

Your task is to implement the solution in code.

The code must:
- Compute the correct number of shares
- Ensure that the attacker cannot buy more than the correct number of shares at the correct price
- Prevent the attacker from stealing the correct amount of W

---
Closes #240
🤖 supa resolutionengine.js


naddison36:
Spam

PR #269, opened 2026-06-11
Make setARMBuffer, setTotalAssetsCap and setLiquidityProviderCaps Talos-ready

clement-ux:
## Summary

Prepares the three remaining permissioned admin tasks (`setARMBuffer`, `setTotalAssetsCap`, `setLiquidityProviderCaps`) so they can be invoked from a Talos schedule command, e.g.:

```
cd /app && pnpm hardhat setTotalAssetsCap --network mainnet --arm Ethena --cap 100000
cd /app && pnpm hardhat setARMBuffer --network mainnet --arm Lido --buffer 0.5
cd /app && pnpm hardhat setLiquidityProviderCaps --network mainnet --arm Ethena --accounts 0x...,0x... --cap 20000
```

### `tasks/admin.js` (purely additive)
- New `setTotalAssetsCap` and `setLiquidityProviderCaps` core functions with an injected signer and the ARM contract as a parameter, following the same pattern as `allocate`/`collectFees`/`setARMBuffer` (and the `action()` wrapper principle on the talos branch).
- `resolveCapManager` helper: resolves the ARM's CapManager via `arm.capManager()` with a clear error when none is configured, using a minimal inline ABI.

### `tasks/liquidityProvider.js`
- The two cap functions become thin Hardhat wrappers (resolve signer + ARM contract by name, then delegate to `admin.js`). Same CLI interface as before. The previous implementation relied on the Hardhat-global `ethers.getContractAt`, which made the logic unusable outside the Hardhat runtime.

### `tasks/tasks.js`
- All params of the three tasks are now **required with no defaults**:
  - `setARMBuffer`: omitting `--buffer` previously wrote a buffer of **0** on-chain silently (`buffer || "0"` fallback); `--arm` defaulted to `Origin`.
  - `setTotalAssetsCap` / `setLiquidityProviderCaps`: `--arm` defaulted to `Lido`.
  - These are punctual state-changing admin actions: a missing flag should fail at CLI parsing (HH306), not write a surprise value to the wrong ARM.
- Param descriptions fixed (liquidity asset units with examples instead of "Amount of WETH").

No defaults is intentional — the Talos actions catalog will report `hasDefault: false` for these flags so the UI knows they are mandatory.

### DIFF
--- a/src/js/tasks/admin.js
+++ b/src/js/tasks/admin.js
@@ -170,8 +170,59 @@ async function setARMBuffer({ arm, signer, buffer }) {
   const tx = await setArmBuffer(arm, signer, bufferBN, { gasLimit });
   await logTxDetails(tx, "setARMBuffer");
 }
+
+const CAP_MANAGER_ABI = [
+  "function setLiquidityProviderCaps(address[] liquidityProviders, uint256 cap)",
+  "function setTotalAssetsCap(uint248 totalAssetsCap)",
+];
+
+// Resolve the CapManager contract of an ARM.
+async function resolveCapManager(arm, signer) {
+  const capManagerAddress = await arm.capManager();
+  if (capManagerAddress === ethers.ZeroAddress) {
+    throw new Error("No CapManager configured for the ARM");
+  }
+  return new ethers.Contract(capManagerAddress, CAP_MANAGER_ABI, signer);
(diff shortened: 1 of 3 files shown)


naddison36 (approved):
LGTM

PR #273, opened 2026-06-17
Talos

clement-ux:
## Summary

Makes the off-chain automation (Hardhat tasks + Defender Actions) work against
**both** the new multi-base ARM and the older **legacy single-asset ARMs** that
are still deployed on-chain, plus a few operational fixes. **No Solidity contract
changes** — this is JS/TS automation, a SQL seed, and a client bump only.

The real diff is 9 files (`git diff main...talos`); see "Why so many commits" below.

### Legacy-aware ARM helpers — `src/js/utils/arm.js`
- **`resolveArmBase`**: fall back to legacy resolution on **any** failure reading
  `baseAssetConfigs`, not just recognizable missing-selector errors. Some RPCs
  surface the absent selector as a bare `execution reverted` with no decodable
  data, which previously threw instead of falling back.
- **`setArmPrices`**: legacy ARMs only expose `setPrices(buyPrice, sellPrice)`
  (2 args). Stop calling the 4-arg multi-base signature on them — the extra
  buy/sell liquidity limits don't exist on legacy contracts.
- **New `getOutstandingWithdrawals(arm)`**: single source of truth for liquidity
  reserved by outstanding LP withdrawals. Tries `withdrawsQueued()/withdrawsClaimed()`,
  falls back to the legacy ABI (the getters were dropped from some new ABIs but
  still exist on the deployed legacy contracts), then to `reservedWithdrawLiquidity()`.
- Restored `withdrawsQueued`/`withdrawsClaimed` to `LEGACY_ARM_ABI`; fixed legacy
  `setPrices` ABI signature to 2 args.

### Consumers use the new helper
- **`tasks/admin.js`** (`collectFees`) and **`tasks/liquidityAutomation.js`**
  (`baseWithdrawAmount`) now call `getOutstandingWithdrawals(arm)` instead of
  reading `withdrawsQueued`/`withdrawsClaimed` directly.

### Ethena unstaker FIFO — `tasks/ethenaQueue.js`
- Stop double-wrapping unstaker entries. Callers already pass `{ address, index }`
  objects via the legacy path; keep them as-is so the real unstaker index is
  preserved for the FIFO `uint8` claim (still tolerates plain address strings).

### Better Action error reporting — `tasks/lib/action.ts`
- Capture and decode on-chain revert data in the action error handler:
  `Error(string)` (`0x08c379a0`), `Panic(uint256)` (`0x4e487b71`), and unknown
  custom selectors. Walks the nested ethers error shape to find `data`.

### Misc
- **`tasks/actions/collectFeesSonic.ts`**: use `OriginARMV1` ABI (was the LidoARM
  ABI) for the Origin ARM on Sonic.
- **`migrations/seed_schedules.sql`**: seed manual-only emergency pause schedules
  for Lido/EtherFi/Ethena (`enabled=false`; triggered via the UI "Run now" button).
- **`package.json` / `pnpm-lock.yaml`**: bump `@talos/client` to `0a6ecc9`.

## Why so many commits for so little change

`main`'s tip (`c0ead70 Talos (#271)`) is itself a **squash-merge** of an earlier
state of this branch, so `main` already contains the bulk of the history. The
branch was then re-synced with `Merge origin/main into talos`, which folded that
squashed snapshot back into the branch's history. Net effect: the ~145 commits
shown collapse to a 9-file diff, because everything before the re-sync is already
on `main`. Only the 5 commits on top of the merge carry new content.

**Recommend "Squash and merge"** so the commit on `main` is exactly these 9 files.


sparrowDom (approved):
LGTM

PR #275, opened 2026-06-18
Add getBaseAssets getter + fix liquidity task queue accounting

clement-ux:
### 1. `getBaseAssets()` getter on `AbstractARM`
Exposes the internal `baseAssets` array via an `external view` getter so off-chain callers can enumerate the supported base assets without reading storage manually.
- NatSpec added, matching the existing getter style (e.g. `asset()`).
- Unit tests cover the empty list, single-asset, and multi-asset (insertion-order preserved) cases.
- Smoke tests: each ARM's `test_initialConfig` asserts `getBaseAssets()` lists the configured base assets (stETH/wstETH, OETH/wOETH, sUSDe, eETH), via a shared membership helper (containment, not exact equality — robust to registration order and future base assets).

### 2. Deploy script `034_UpgradeEthenaARMGetBaseAssetsScript`
The deployed Ethena ARM runs the `031` implementation, which predates `getBaseAssets()`. Unlike Lido/EtherFi/Origin, the smoke harness does not redeploy the Ethena logic when `031` is already live on the fork, so the getter reverted under CI (which forks at `latest`).
- Logic-only upgrade: deploys a fresh `EthenaARM` implementation and upgrades the multisig-owned proxy to it.
- Storage layout unchanged; sUSDe stays registered from `031` (no adapter redeploy / base-asset registration).
- The smoke harness now upgrades the Ethena ARM to this implementation, so `getBaseAssets()` is available and the assertion passes; on mainnet it brings the deployed ARM up to this version.

### 3. Ethena claim smoke test made fork-state independent
`EthenaAssetAdapter.redeem()` is strictly FIFO and requires the claimed shares to match the front request. On a `latest` fork the live adapter can already hold real, in-flight sUSDe cooldown requests at the head of its queue, so `test_claim_ethena_request_with_delay` reverted with `Adapter: invalid redeem amount`. The test now advances the adapter's claim cursor (`nextPendingIndex = totalRequests`) past any pre-existing requests before making its own, so it passes regardless of live on-chain queue state. A storage-layout guard fails loudly if the adapter's slots are ever reordered.

### 4. `liquidity.js` outstanding-queue fix
Since the withdrawal queue refactor (#208) the queue is tracked in **shares**, so the old `withdrawsQueued - withdrawsClaimed` no longer reflects the outstanding obligation in liquidity-asset terms. The task now reads `reservedWithdrawLiquidity` directly.

## Testing
- `forge build` — passes (pre-existing warnings only)
- `forge test --match-path "test/unit/**"` — 249 passed, 0 failed
- `make test-smoke` (pinned fork block) — 49 passed, 0 failed
- Full smoke suite at `latest` fork block (CI condition) — all ARMs green, incl. Ethena 12/12
- `forge fmt --check` — clean; `prettier`/`eslint` on `liquidity.js` — clean

### DIFF
--- a/script/deploy/mainnet/034_UpgradeEthenaARMGetBaseAssetsScript.s.sol
+++ b/script/deploy/mainnet/034_UpgradeEthenaARMGetBaseAssetsScript.s.sol
@@ -1,0 +1,44 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.23;
+
+// Contract
+import {Proxy} from "contracts/Proxy.sol";
+import {Mainnet} from "contracts/utils/Addresses.sol";
+import {EthenaARM} from "contracts/EthenaARM.sol";
+
+// Deployment
+import {AbstractDeployScript} from "script/deploy/helpers/AbstractDeployScript.s.sol";
+
+/// @title Upgrade Ethena ARM to expose getBaseAssets()
+/// @notice Upgrades the multisig-owned Ethena ARM to a fresh EthenaARM implementation that exposes
+///         the new `getBaseAssets()` getter added to AbstractARM. The 031 implementation predates the
+///         getter, so this is a logic-only upgrade: storage layout is unchanged, and no adapter
+///         redeployment or base-asset registration is needed (sUSDe stays registered from 031).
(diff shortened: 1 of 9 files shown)


naddison36 (approved):
LGTM

PR #281, opened 2026-06-25
[ARM] Ether ARM

clement-ux:
Summary:
- Add fresh-deploy EtherARM.
- Remove legacy upgrade storage from AbstractARM and pack withdrawal requests by deriving shares from cumulative queued shares.
- Expose withdrawalRequestShares() and update tests/helpers for the new withdrawalRequests() shape.

Reason:
- The ETH ARM will be deployed from scratch, so legacy upgrade storage is no longer needed.
- Deriving request shares saves one storage slot per withdrawal request while keeping loss socialization through escrowed shares.

### DIFF
--- a/src/contracts/AbstractARM.sol
+++ b/src/contracts/AbstractARM.sol
@@ -13,8 +13,8 @@ import {IAssetAdapter, IERC20, ICapManager} from "./Interfaces.sol";
  * @title Generic Automated Redemption Manager (ARM)
  * @notice Coordinates liquidity-provider shares, two-token swaps, active market allocation, and
  * protocol-specific redemption adapters for one liquidity asset and one or more supported base assets.
- * @dev Existing ARM proxies depend on the original storage prefix. New multi-base state is appended after
- * legacy single-base storage so Lido, EtherFi, Ethena, and Origin ARMs can share this implementation.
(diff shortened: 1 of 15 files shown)


src/contracts/AbstractARM.sol
```diff
@@ -50,22 +50,8 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable, ReentrancyGu
...
 
     ////////////////////////////////////////////////////
-    ///                 Storage
+    ///                 Structs
     ////////////////////////////////////////////////////
-
-    /// @dev Legacy single-base storage. Keep this prefix unchanged for existing proxy upgrades.
-    /// These fields are retained for storage/ABI compatibility and are not the source of truth for
-    /// multi-base swap pricing.
-    uint256 internal _deprecatedTraderate0;
```
naddison36, line 59: I think we should keep these for now as it gives us the option to upgrade the Lido and EtherFi ARMs if we change our minds

naddison36 (commented):
I think we should keep the deprecated storage slots for now just in case we change our minds and want to upgrade the existing EtherFi and Lido ARMs

clement-ux:
Closed, in favor of this one: https://github.com/OriginProtocol/arm-oeth/pull/282

PR #294, opened 2026-07-08
Cleanup addresses and rename ETH ARM

clement-ux:
## Description
cleanup address et rename Eth ARM

## Testing
- forge build


naddison36 (approved):
Approved

PR #292, opened 2026-07-08
[ARM] Deploy 037 USD ARM (PYUSD/USDG) on mainnet

clement-ux:
# Description

Record the mainnet deployment of the **USD ARM**: a USDC-quoted `MultiAssetARM` with two Paxos-issued base assets, each wired through its own `PaxosAssetAdapter` behind a proxy:

- **PYUSD** — pegged 1:1 to USDC, redeemed off-chain through Paxos Actions that settle USDC back to the adapter.
- **USDG** — pegged 1:1 to USDC, same off-chain Paxos redemption flow.

Ownership of the ARM, its `CapManager` and both adapter proxies is handed to the multi-chain 2/8 multisig (`STRATEGIST`); the operational role (request/claim redemptions, submit Paxos redeems, set prices) is the Talos KMS relayer (`ARM_TALOS_RELAYER`). No lending market is wired up — idle USDC stays in the ARM until a market is added in a follow-up script.

This is a rename of the former `037_DeployPaxosARMScript`: only the contract name, the recorded deployment names (`STABLES_ARM*` → `USD_ARM*`) and the ARM name/symbol (`Stables ARM` / `ARM-USDC-Stables` → `USD ARM` / `ARM-USD`) changed; the deployment logic is identical. It supersedes #290.

> The companion WETH-quoted **LST ARM** (`036_DeployMultiAssetARMScript`) deployed in the same round is recorded in a separate PR.

The following PRs are included in this deployment

### ARM

- https://github.com/OriginProtocol/arm-oeth/pull/289

## Deployment

Script `script/deploy/mainnet/037_DeployUSDARMScript.s.sol`

## Contracts

| Contract | Address |
| - | - |
| USD ARM (`MultiAssetARM`) proxy | [0x5c68a604d39A639C0b8CC115f0823C50272CF13a](https://etherscan.io/address/0x5c68a604d39A639C0b8CC115f0823C50272CF13a) |
| USD ARM implementation (`MultiAssetARM`) | [0x1405CDB24f93A60d023400cb03787bB8576EE490](https://etherscan.io/address/0x1405CDB24f93A60d023400cb03787bB8576EE490) |
| USD ARM `CapManager` proxy | [0xAAB4F9e842e5D34ec61B1DBAAE8EBa4559F0eEF6](https://etherscan.io/address/0xAAB4F9e842e5D34ec61B1DBAAE8EBa4559F0eEF6) |
[...]

PR #295, opened 2026-07-08
Add Talos actions for the USD ARM (PYUSD/USDG)

clement-ux:
# Description

Adds the Talos off-chain automation for the **USD ARM** (`MultiAssetARM`, USDC-quoted, PYUSD/USDG base assets) deployed in #292.

## Dedicated actions (`src/js/tasks/actions/`)

| Action | Behaviour |
| - | - |
| `setPricesUSD` | Kyber quotes for PYUSD/USDC and USDG/USDC via `setPricesForBases`, bounded by `--min/max-buy/sell-price` (defaults: buy in [0.998, 0.99995], sell in [0.99997, 1.0], cross price 0.99997). Quote amount defaults to the ARM's withdrawable USDC reserves; **skips while less than 100 USDC is withdrawable** (dust amounts get quantized into garbage price ratios and Kyber rejects them, so the action would error while the ARM is unfunded). `--bases` targets a single asset. Swap caps are reset to `uint128.max` on each update (same convention as Ethena/Lido) — **the first successful run therefore enables swaps** (deployed with 0/0). |
| `autoRequestUSDWithdraw` | Per base (`--bases`, default `PYUSD,USDG`): `requestBaseAssetRedeem` when the ARM balance is above `--threshold-amount` (default 1000), then `submitPaxosRedeem(pendingShares, id)` on the adapter with an off-chain generated reconciliation id (eg `PYUSD-1783500329`). Submitting all pending shares also flushes shares left by an earlier failed submit. **Refuses to submit while `paxosRecipient` is still the `0x…dEaD` placeholder.** |
| `autoClaimUSDWithdraw` | Per base (`--bases`): claims `min(settlingShares, USDC settled on the adapter)` back into the ARM once above `--min-amount` (default 100), so partial Paxos settlements are claimed incrementally. |
| `allocateUSD` | Standard `allocate` with a 6-decimals threshold. Exits cleanly while no lending market is wired (`activeMarket == 0`). |
| `collectUSDFees` / `pauseUSD` | Same as the Ethena equivalents. |

The Paxos request/claim logic lives in `src/js/tasks/paxosQueue.js` (analog of `ethenaQueue.js`, no cooldown/FIFO handling needed — the adapter is 1:1 with no request delay).

## General mainnet actions

`setARMBufferAction`, `setLiquidityProviderCapsAction` and `setTotalAssetsCapAction` learn the new `--arm usd` key via `MAINNET_ARMS` (`src/js/tasks/lib/arm.ts`), which now carries the liquidity-asset `decimals` per ARM.

## Decimals handling

The shared helpers assumed 18 decimals everywhere (`parseUnits` with no unit). Two fixes, both defaulting to 18 so existing ARMs are bit-for-bit unchanged:

**Explicit liquidity-asset decimals** for the USDC-quoted ARM:
- `admin.js`: `setTotalAssetsCap` / `setLiquidityProviderCaps` (a `--cap 100000` would have written `1e23` instead of `100_000e6`), `allocate` threshold and dust check, `collectFees` logging

**Mixed base/liquidity decimals** — an 18-dec base over the 6-dec USDC liquidity asset is valid on-chain (only the current `PaxosAssetAdapter` enforces equality), so base and liquidity decimals are threaded separately:
- `kyber.js` / `1Inch.js`: each swap leg scales its `amountIn` with its own token decimals (liquid for the buy leg, base for the sell leg) and the price ratios are normalized to 1e18
- `armPrices.js`: base decimals come from the on-chain `baseAssetConfigs` (new `baseAssetDecimals` field, absent on legacy ABIs → 18); liquidity decimals are read on-chain from the token; both are threaded through the aggregator calls and the wrapped-asset rate adjustment
- `liquidityAutomation.js` `baseWithdrawAmount`: separate `liquidityDecimals` option for the ARM/lending-market liquidity amounts and comparisons
- `paxosQueue.js` claim: the settled liquidity balance is scaled to share units before `min(settlingShares, settled)`

`src/abis/MultiAssetARM.json` is added from the forge artifact: the existing `EthenaARM.json` ABI has the old 8-field `baseAssetConfigs` tuple (no `baseAssetDecimals`) and would silently decode a garbage adapter address on the new ARM.

## Schedules

Six `mainnet_*_usd` rows added to `migrations/seed_schedules.sql` (all `enabled=false`, crons staggered against the existing ARMs to avoid relayer-nonce and Kyber rate-limit collisions, pause row in the manual-only block). The three general admin rows are shared across ARMs via `--arm` and need no new rows. Deploying still requires applying the seed to the shared Talos Postgres and rebuilding the Docker image.

## Testing

Validated on anvil mainnet forks (`IMPERSONATE` = Talos relayer), including a full Paxos round-trip with real funds:
1. 5,000 PYUSD seeded into the ARM → `autoRequestUSDWithdraw` executed `requestBaseAssetRedeem` then **failed loudly on the placeholder-recipient guard** (by design)
2. `setPaxosRecipient` set (impersonated 2/8 owner) → re-run submitted all 5,000 pending shares to Paxos with a readable redemption id
3. 5,000 USDC transferred to the adapter (simulated Paxos settlement) → `autoClaimUSDWithdraw` claimed it back into the ARM, `settlingShares` → 0
4. `setPricesUSD` skips cleanly on the unfunded ARM; with reserves (and with `--amount 5000`) it fetched live Kyber quotes for both pairs (a USDG↔USDC route exists — PYUSD mid 1.0003, USDG mid 0.99999) and set prices on-chain with `uint128.max` swap caps
5. `--bases PYUSD` / `--bases USDG` runs touch only the requested asset
6. `setTotalAssetsCapAction --arm usd --cap 100000` → `100_000e6` on the CapManager (also LP caps and ARM buffer)
7. `allocateUSD` (skips, no market), `collectUSDFees`, `pauseUSD` all clean
8. Regressions: `setPricesEthena` (incl. the wrapped sUSDe rate adjustment) and `autoRequestEthenaWithdraw` unchanged at 18 decimals; the normalized aggregator ratio math is bit-for-bit identical to the old formula for equal-decimals pairs

`pnpm lint`, `pnpm typecheck` and `pnpm prettier:check` pass; all six tasks and their flags appear in `dump-actions-catalog.cjs`.

## Before enabling the schedules

- The 2/8 multisig must call `setPaxosRecipient()` on **both** adapters (deployed with the `0x…dEaD` placeholder) before enabling `mainnet_auto_request_usd_withdraw`.
- The first successful `setPricesUSD` run enables swaps (caps 0 → unlimited).
- The `--fee` half-spread default (2 bps) mirrors Ethena and can be tightened from the Talos UI for a stable/stable pair.


src/js/tasks/actions/pauseUSD.ts
```diff
@@ -0,0 +1,18 @@
+import { ethers } from "ethers";
+
+import { action } from "../lib/action";
+import { pauseARM } from "../admin";
+import { mainnet } from "../../utils/addresses";
+const multiAssetARMAbi = require("../../../abis/MultiAssetARM.json");
+
+action({
+  name: "pauseUSD",
```
naddison36: I've prefer to consolidated all the pause actions into one like `setARMBufferAction`. That is, one `pause` actions with a `--arm` param.
But happy to keep the separate `pauseUSD` Action for now

naddison36: I'll so this change in a new PR to this branch

naddison36 (dismissed):

src/js/tasks/actions/pauseUSD.ts
clement-ux: I had thought of that too—thanks for doing it!

naddison36 (approved):
Looks good

PR #298, opened 2026-07-10
Fix Ethena unstaker FIFO ordering after rotation

clement-ux:
## Summary

- query each current Ethena unstaker slot once instead of expanding every historical request
- reconstruct the pending FIFO from `totalRequests`, the active request count, and `MAX_UNSTAKERS`
- add regression coverage for the first 42-slot wrap and duplicate-share prevention

## Root cause

After `totalRequests` exceeded 42, the action expanded every historical request id and resolved each id modulo 42. At 45 requests this produced `0, 1, 2, ..., 41, 0, 1, 2`, while the actual pending FIFO was `41, 0, 1, 2`. Since slots 0-2 were not ready, the off-chain FIFO prefix stopped before ready slot 41 and submitted no claim.

## Mainnet validation

Against the current Ethena adapter state:

- `totalRequests = 45`
- reconstructed pending order: `[41, 0, 1, 2]`
- reconstructed claimable prefix: `[41]`
- claim shares: `6063070730252694374950`

A read-only simulation from the Talos operator returns `7509122412196736546883` USDe for that claim.

## Tests

- `node test/js/ethenaQueue.test.js`
- `pnpm lint`
- `forge test --match-path test/unit/adapters/concrete/EthenaAssetAdapter.t.sol` (21 passed)
- focused Prettier check on changed JS files

`pnpm typecheck` remains blocked by the existing repository setup: `tsconfig.json` requests the Node type library, but `@types/node` is not declared as a root dependency.

### DIFF
--- a/src/js/tasks/ethenaQueue.js
+++ b/src/js/tasks/ethenaQueue.js
@@ -8,6 +8,10 @@ const {
   requestBaseAssetWithdrawal,
   resolveArmBase,
 } = require("../utils/arm");
+const {
+  orderPendingUnstakerStates,
+  selectClaimableFifoPrefix,
+} = require("../utils/ethenaQueue");
 const { logTxDetails } = require("../utils/txLogger");
 const log = require("../utils/logger")("task:ethenaQueue");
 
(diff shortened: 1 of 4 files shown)


naddison36 (approved):
LGTM

PR #306, opened 2026-07-21
[yAudit-03] Fix swap fee accrual on realized gain

clement-ux:
## Summary

- accrue buy-side performance fees from the value of the actual settled input rather than the quoted notional spread
- pass amountIn through swap liquidity validation so exact-input and exact-output swaps use the same realized-gain basis
- update fork, fuzz, and invariant fee expectations and add a regression test for 18-to-6 decimal truncation below one base unit
- harden the Ethena and Lido invariant handlers so corpus mutations remain ABI-decodable in strict mode

## Invariant harness

The invariant handlers accept full-word uint256 inputs and derive booleans or bound narrower values internally. This prevents mutated calldata from reverting in the Solidity ABI dispatcher before handler execution.

## Testing

- Full suite excluding invariants: 531 passed, 0 failed, 3 skipped
- Ethena invariants with FOUNDRY_INVARIANT_FAIL_ON_REVERT=true: 256 runs, 12,800 calls, 0 reverts
- Lido invariants with FOUNDRY_INVARIANT_FAIL_ON_REVERT=true: 256 runs, 12,800 calls, 0 reverts

## Audit report

Related yAudit issue: https://github.com/yAudit/origin-paxos-adapter-review/issues/3

### DIFF
--- a/src/contracts/AbstractARM.sol
+++ b/src/contracts/AbstractARM.sol
@@ -428,7 +428,7 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable, ReentrancyGu
             amountOut = convertedAmountIn * config.buyPrice / PRICE_SCALE;
         }
 
-        _validateAndConsumeSwapLiquidity(config, isBuySide, outToken, amountOut);
+        _validateAndConsumeSwapLiquidity(config, isBuySide, outToken, amountIn, amountOut);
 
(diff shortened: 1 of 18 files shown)


naddison36 (approved):
This is a good find.
I've checked the new method is economically equivalent without rounding issues. The new method is better as it handles the rounding.

PR #308, opened 2026-07-21
[yAudit-02] Allow deposits at exact solvency

clement-ux:
## Summary

- treat exact coverage of accrued fees plus `MIN_LIQUIDITY` as solvent
- keep deposits blocked when gross assets are even one unit below the required floor
- add regression coverage for both 6-decimal and 18-decimal configurations

Fixes the finding reported in https://github.com/yAudit/origin-paxos-adapter-review/issues/2

## Testing

- `forge test --match-path 'test/unit/MultiAssetARM/concrete/Deposit.t.sol'` (16 passed)
- `forge test --no-match-path 'test/invariants/**'` (532 passed, 0 failed, 3 skipped)


naddison36 (approved):
ok

PR #311, opened 2026-07-21
Block deposits at the asset floor when live LPs exist

clement-ux:
## Summary

- block deposits at the native-liquidity floor when live LP shares still exist
- preserve the initial-deposit path when dead shares are the entire supply and no senior liabilities exist
- add regression coverage for a real asset loss with zero accrued fees and zero reserved withdrawals
- align the Ethena and Lido invariant deposit guards with the new insolvency condition
- harden Lido invariant handlers so mutated calldata remains ABI-decodable and strict mode runs with zero handler reverts

## Security impact

A real market loss can reduce total assets to the MIN_LIQUIDITY floor without creating fees or reserved withdrawals. Before this change, a new depositor could mint most of the share supply at the floor price and capture a later recovery that belongs to existing LPs.

This is a follow-up to #288: that PR covered fee-driven insolvency, while this PR covers the separate asset-loss case.

## Invariant harness

The Lido handlers accept full-word uint256 seeds and bound or derive narrower values internally. This prevents corpus mutations from producing non-canonical bool or narrow-integer ABI words that revert in the Solidity ABI dispatcher before handler execution. The handlers also reject zero-input dust cases and safely normalize full-width seeds and uint128 price-liquidity limits.

## Testing

- Deposit tests: 16 passed, 0 failed
- Full suite excluding invariants: 532 passed, 0 failed, 3 skipped
- Ethena invariants with FOUNDRY_INVARIANT_FAIL_ON_REVERT=true: 256 runs, 12,800 calls, 0 reverts
- Lido invariants with FOUNDRY_INVARIANT_FAIL_ON_REVERT=true: 256 runs, 12,800 calls, 0 reverts

### DIFF
--- a/src/contracts/AbstractARM.sol
+++ b/src/contracts/AbstractARM.sol
@@ -783,11 +783,12 @@ abstract contract AbstractARM is OwnableOperable, ERC20Upgradeable, ReentrancyGu
         uint256 grossAssets = _availableAssets();
         uint256 feesAccruedMem = feesAccrued;
 
-        // Treat accrued fees as a senior liability. If gross assets no longer cover
-        // accrued fees plus the minimum native-liquidity floor, new deposits would be
-        // minted against the floor and backfill the shortfall, so block deposits.
+        // At the native-liquidity floor, a new deposit would either backfill senior liabilities
+        // or dilute live LPs after a real asset loss. Allow only the initial deposit, when the dead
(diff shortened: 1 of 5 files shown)


naddison36 (dismissed):
Another good find

naddison36 (approved):
post merge approve

PR #313, opened 2026-07-22
fix(invariants): account for ERC4626 market rounding

clement-ux:
## Summary

- replace the fixed 100 wei WETH conservation tolerance with exact ghost accounting for ERC4626 market rounding
- bound every recorded loss to at most one market share value so unexpected losses still fail the invariant
- cover handlers that can deposit into, withdraw from, or revalue the ARM market position
- add a deterministic regression for the CI counterexample: 101 wei of cumulative rounding after 76 market rotations

## Root cause

The failing LidoARM invariant alternates the active market 76 times. Each ERC4626 deposit rounds minted shares down and can lose up to two wei at the mocked market exchange rates. The cumulative loss is 101 wei, which exceeds the previous fixed 100 wei tolerance by one wei.

This is expected bounded ERC4626 rounding rather than an unexplained balance loss. A fixed tolerance is not invariant-safe because the loss grows with the number of market interactions.

Failing job: https://github.com/OriginProtocol/arm-oeth/actions/runs/29904352973/job/88872109503

## Testing

- `forge build`
- `forge fmt --check`
- `forge test --summary --fail-fast --show-progress --mp 'test/unit/**'`
- `FOUNDRY_PROFILE=lite FOUNDRY_INVARIANT_FAIL_ON_REVERT=true forge test --match-contract FuzzerFoundry_LidoARM -vv`
- `forge test --match-contract FuzzerFoundry_LidoARM --fuzz-seed 0xcd574f81fe803283b87931d638f92df2698e5688c663d73b5d0bb1865ecb2161 -vv`

No production contract code is changed.

### DIFF
--- a/test/invariants/LidoARM/FuzzerFoundry.t.sol
+++ b/test/invariants/LidoARM/FuzzerFoundry.t.sol
@@ -101,6 +101,25 @@ contract FuzzerFoundry_LidoARM is Properties {
         require(property_bal_wsteth(), "BAL_WSTETH: wstETH balance mismatch");
     }
 
+    function test_marketRotationRoundingLossIsAccountedFor() public {
+        targetDeposit(23_061, 1);
+        targetRequestRedeem(14_558, 1);
+
+        for (uint256 i; i < 36; ++i) {
+            targetSetActiveMarket(1);
+        }
+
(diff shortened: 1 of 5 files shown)


naddison36 (approved):
LGTM

PR #320, opened 2026-07-24
Etherfi Withdrawal Request Gate

sparrowDom:
## Summary

Gates `EtherFiAssetAdapter` & `WeETHAssetAdapter` so an Ether.fi withdrawal NFT owned by the adapter can only be claimed *through the adapter itself*.

## Background

Ether.fi's `WithdrawRequestNFT.claimWithdraw(tokenId)` is **permissionless**: anyone can claim a finalized withdrawal NFT, and the proceeds are sent to the NFT owner. In the multi-asset ARM the NFT is owned by `EtherFiAssetAdapter` or `WeETHAssetAdapter`, the ARM values the outstanding request once via `pendingRedeemAssets`, and the adapter's `redeem()` wraps and sweeps its **entire** ETH/WETH balance to the ARM.

## Fix

A reentrancy-style flag marks the only legitimate window in which Ether.fi may pay this adapter, and `receive()` rejects claim proceeds that arrive outside it

### DIFF
--- a/src/contracts/adapters/EtherFiAssetAdapter.sol
+++ b/src/contracts/adapters/EtherFiAssetAdapter.sol
@@ -166,11 +166,11 @@ contract EtherFiAssetAdapter is Initializable, IAssetAdapter, IERC721Receiver {
         return pendingRequestIds[index];
     }
 
-    /// @notice Accepts ETH donations and authorized Ether.fi claim proceeds.
-    /// @dev A permissionless Ether.fi claim is rejected unless it was initiated by this adapter.
-    /// Ether.fi reverts the entire claim, including the NFT burn, when this transfer fails.
+    /// @notice Accepts ETH only while this adapter is claiming Ether.fi withdrawals.
+    /// @dev ETH arriving outside an adapter-initiated claim (e.g. a permissionless Ether.fi claim) is
+    /// rejected. Ether.fi reverts the entire claim, including the NFT burn, when this transfer fails.
     receive() external payable {
-        if (msg.sender == address(etherfiWithdrawalNFT) && !claimingEtherFi) {
+        if (!claimingEtherFi) {
             revert UnauthorizedEtherFiClaim();
         }
     }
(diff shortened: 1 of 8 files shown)


src/contracts/adapters/EtherFiAssetAdapter.sol
```diff
@@ -158,7 +166,14 @@ contract EtherFiAssetAdapter is Initializable, IAssetAdapter, IERC721Receiver {
         return pendingRequestIds[index];
     }
 
-    receive() external payable {}
+    /// @notice Accepts ETH only while this adapter is claiming Ether.fi withdrawals.
+    /// @dev ETH arriving outside an adapter-initiated claim (e.g. a permissionless Ether.fi claim) is
+    /// rejected. Ether.fi reverts the entire claim, including the NFT burn, when this transfer fails.
+    receive() external payable {
+        if (!claimingEtherFi) {
```
sparrowDom, line 173: @clement-ux is making a point that this check could be: 
```
 if (msg.sender == address(etherfiWithdrawalNFT) && !claimingEtherFi) {
```
There are pros/cons to it: 
🟢 code change would allow sending ETH to adapter in case there was a shortfall
🔴 code change would allow a donation attack

sparrowDom: I lean towards not allowing for a donation attack and accounting for everything. In case we ever need to infuse the adapter/ARM with funds I would rather we make a deliberate function for it

pandadefi: You can always send weth

clement-ux: After some thought, I agree that we should keep the first implementation `(msg.sender == address(etherfiWithdrawalNFT) && !claimingEtherFi)`.

shahthepro (dismissed):

naddison36 (dismissed):
LGTM

naddison36 (approved):
LGTM

src/contracts/adapters/EtherFiAssetAdapter.sol
sparrowDom: nit: you probably ment this by the first implementation: 
```
if (!claimingEtherFi) {
```

PR #345, opened 2026-08-31
Pin Foundry to v1.8.1

clement-ux:
## Summary

- Pin Foundry v1.8.1 for local installation and GitHub Actions.
- Pin the Docker Foundry stage to the immutable v1.8.1 multi-architecture digest.
- Correct the foundryup install command and align the architecture documentation.

## Verification

- `git diff --check`
- Parsed `.github/actions/setup/action.yml` as YAML
- `make -n install`
- Verified the v1.8.1 manifest digest and image metadata against GHCR


naddison36 (approved):
LGTM

